Issue #13, opened 2021-08-11
Interface with league ref application  / auto-refs


JNeiger:
As a application, our software must be able to be controlled by the league referee applications/autoref. Interfacing with the outside protocol and converting it to some internal message is critical.

PR #36, opened 2021-08-18
Vision and autoref bridge

### LINKED ISSUE #15: Interface with SSL-Vision
It is required that we can take vision measurements from SSL-Vision and convert them into an internal type.

JNeiger:
Just wanted to throw this here to get some initial feedback about how I'm setting up the ROS packages. Still doesn't compile, but I can fix that later

- Moved protobufs to their own package so we can use them in the ref and vision package
- Moved multicast receiver into common package so we don't duplicate the code
- Started adding vision forwarding that matches the referee bridge style of implementation
- Added ref bridge

Closes #15 #13


barulicm:
I like it. The packages you've split off make a lot of sense. 👍

JNeiger:
Should be good to go. I'm seeing a few messages a second that aren't being parsed correctly, but all the data seems valid so far.

Like the autoref, I can't seem to get multicast working with docker grsim, only running locally. I'm using WSL so it may be that?

Also, this is diff'd onto `dev/barulicm/autoref_bridge`. We'll need to change it once that gets merged.

JNeiger:
Per slack, we're just going to merge this in and have vision and referee stuff in the same PR. All is tested locally and works on my computer

.github/workflows/CI.yml
```diff
@@ -18,6 +18,7 @@ jobs:
         run: |
           source /opt/ros/galactic/setup.bash
           sudo apt-get update
+          sudo apt-get install -y protobuf-compiler
```
barulicm, line 21: This can be installed by adding `protobuf-dev` as a dependency to the package that needs it. (This is the rosdep key that includes the `protobuf-compiler` apt package).

ateam_common/src/multicast_receiver.cpp
```diff
@@ -0,0 +1,53 @@
...
+
+void MulticastReceiver::HandleMulticastReceiveFrom(const boost::system::error_code& error, size_t bytes_received)
+{
+  if(error) {
+    std::cerr << "Receive from error: " << error.message() << std::endl;
+    return;
+  }
+
+  if(!receive_callback_(buffer_.data(), bytes_received)) {
+    std::cerr << "Failed to parse message." << std::endl;
```
barulicm, line 53: Should we move this logging into the callback itself, and not need it to return anything? Could also make it easier to send this message to the ROS log.

ssl_league_msgs/msg/VisionDetectionRobot.msg
```diff
@@ -0,0 +1,6 @@
+float32 confidence
+uint32 robot_id
+geometry_msgs/Point32 pos
```
barulicm, line 3: Thoughts on combining `pos` and `orientation` using `geometry_msgs/msg/Pose2D`? 
https://docs.ros2.org/galactic/api/geometry_msgs/msg/Pose2D.html

ssl_league_msgs/msg/VisionGeometryCameraCalibration.msg
```diff
@@ -0,0 +1,7 @@
+uint32 camera_id
+float32 focal_length
+geometry_msgs/Point32 principal_point
+float32 distortion
+geometry_msgs/Quaternion q
```
barulicm, line 5: Thoughts on combining `q` and `t` using `geometry_msgs/msg/Pose`?
https://docs.ros2.org/galactic/api/geometry_msgs/msg/Pose.html

barulicm (commented):
Lookin good! Just a couple comments / questions.

ateam_common/src/multicast_receiver.cpp
JNeiger: That makes a lot of sense, done

ssl_league_msgs/msg/VisionDetectionRobot.msg
JNeiger: Pose2D looks to be depreceated in Foxy, but I couldn't figure out a one liner to do yaw->quat so I'm cool just doing Pose2D until it finally dies and we swap.

barulicm: Good catch, I completely missed that. 
One way to do yaw->quat is to use the utilities in `tf2_geometry_msgs`:
```
auto quat_msg = tf2::toMsg(tf2::Quaternion(tf2::Vector3(0,0,1), yaw));
```

ateam_autoref_bridge/src/autoref_bridge_node.cpp
```diff
@@ -18,12 +18,11 @@ class AutorefBridgeNode : public rclcpp::Node
       10003,
       [this](auto * buffer, size_t bytes_received) {
         Referee referee_proto;
-        if (!referee_proto.ParseFromArray(buffer, bytes_received)) {
-          return false;
+        if (referee_proto.ParseFromArray(buffer, bytes_received)) {
+          referee_publisher_->publish(message_conversions::fromProto(referee_proto));
+        } else {
+          RCLCPP_INFO(get_logger(), "Failed to parse referee protobuf packet");
```
barulicm, line 24: nit: should probably be `RCLCPP_WARN` or `RCLCPP_ERROR`.

JNeiger:
Ayyy, got CI to pass, we good to go pending reviews

barulicm (approved):

PR #38, opened 2021-09-02
Robot Control -> Simulation bridge

### LINKED ISSUE #2: Bi-directional radio communication with sim (grSim)
Communicate robot commands and feedback to and from grSim

JNeiger:
Closes #2 

Setups up a ROS msg -> protobuf for robot control in simulation. Just using the default world vel command for now, but this in no way is a solid call, just a placeholder for now

### DIFF
--- a/ateam_common/include/ateam_common/udp_sender.hpp
+++ b/ateam_common/include/ateam_common/udp_sender.hpp
@@ -18,8 +18,8 @@
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 // THE SOFTWARE.
 
-#ifndef ATEAM_COMMON__MULTICAST_SENDER_HPP_
-#define ATEAM_COMMON__MULTICAST_SENDER_HPP_
+#ifndef ATEAM_COMMON__UDP_SENDER_HPP_
+#define ATEAM_COMMON__UDP_SENDER_HPP_
 
 #include <boost/asio.hpp>
 
@@ -28,27 +28,27 @@
 
 namespace ateam_common
 {
-class MulticastSender
+class UDPSender
 {
 public:
-  MulticastSender(
-    std::string multicast_ip_address,
-    int16_t multicast_port);
+  UDPSender(
+    std::string udp_ip_address,
+    int16_t udp_port);
 
   void send(char * data, size_t length);
 
-  ~MulticastSender();
+  ~UDPSender();
 
 private:
   boost::asio::io_service io_service_;
-  boost::asio::ip::udp::socket multicast_socket_;
+  boost::asio::ip::udp::socket udp_socket_;
   boost::asio::ip::udp::endpoint receiver_endpoint_;
   std::array<char, 1024> buffer_;
   std::thread io_service_thread_;
 
-  void HandleMulticastSendTo(const boost::system::error_code& error, std::size_t bytes_transferred);
+  void HandleUDPSendTo(const boost::system::error_code& error, std::size_t bytes_transferred);
 };
 
 }  // namespace ateam_common
 
-#endif  // ATEAM_COMMON__MULTICAST_SENDER_HPP_
+#endif  // ATEAM_COMMON__UDP_SENDER_HPP_
--- a/ateam_common/include/ateam_common/udp_sender.hpp
+++ b/ateam_common/include/ateam_common/udp_sender.hpp
@@ -32,10 +32,10 @@ class UDPSender
 {
 public:
   UDPSender(
-    std::string udp_ip_address,
-    int16_t udp_port);
+    const std::string & udp_ip_address,
+    const int16_t udp_port);
 
-  void send(char * data, size_t length);
+  void send(const char * const data, const size_t length);
 
   ~UDPSender();
 
--- a/ateam_common/include/ateam_common/udp_sender.hpp
+++ b/ateam_common/include/ateam_common/udp_sender.hpp
@@ -46,7 +46,7 @@ private:
   std::array<char, 1024> buffer_;
   std::thread io_service_thread_;
 
-  void HandleUDPSendTo(const boost::system::error_code& error, std::size_t bytes_transferred);
+  void HandleUDPSendTo(const boost::system::error_code & error, std::size_t bytes_transferred);
 };
 
 }  // namespace ateam_common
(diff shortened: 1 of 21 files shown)


JNeiger:
Good to go!

To run, execute the following, each in a new terminal

- ros2 run ateam_ssl_simulation_radio_bridge ssl_simulation_radio_bridge_node
- ./grSim --headless -platform offscreen
  - Need to clone + build grsim, execute in ~/grSim/bin
- yarn serve
  - Need to clone + build ssl-vision-client
- go run cmd/ssl-vision-client/main.go
  - Need to clone + build ssl-vision-client
  - Change port to  60020 in main.go for vision (grSim defaults to 60020 instead of 60003 for some reason)
- ros2 topic pub /ateam_ssl_simulation_radio_bridge/robot_commands ateam_msgs/msg/RobotCommands "{ commands: [ {id: 3, twist: { linear: { x: 5, y: 5, z: 0}, angular: { x: 0, y: 0, z: 1} } }]}"
- Open http://localhost:8080/ in chrome and see blue team of id 3 move around

barulicm:
I think there's an interesting conversation from this for us to maybe have at the next meeting:
Where do we want to put the robot id?

- In the messages
  - As done currently in this PR every command or sensor reading to/from the robots comes across one topic and all messages have to have a robot ID field
- In the topic name
  - Use topic names to isolate command / sensor streams for each robot. Ie. `/robot/0/motion_commands` using standard geometry_msgs/Twist messages.

JNeiger:
Fixed to have a separate topic for each robot. Message can probably just be simplified to a twist, but leaving it for now so we don't have to redo work to create this message package. I imagine it'll actually change in the future anyways

Note: Can't actually start a topic with a number so doing `robot#` instead
```
joe@DESKTOP-2H0D5SC:~$ ros2 run ateam_ssl_simulation_radio_bridge ssl_simulation_radio_bridge_node
terminate called after throwing an instance of 'rclcpp::exceptions::InvalidTopicNameError'
  what():  Invalid topic name: topic name token must not start with a number:
  '~/robot_commands/0'
                    ^

joe@DESKTOP-2H0D5SC:~$ ros2 run ateam_ssl_simulation_radio_bridge ssl_simulation_radio_bridge_node
terminate called after throwing an instance of 'rclcpp::exceptions::InvalidTopicNameError'
  what():  Invalid topic name: topic name token must not start with a number:
  '~/0/robot_motion_command'
     ^
```

ateam_common/include/ateam_common/udp_sender.hpp
```diff
@@ -0,0 +1,54 @@
...
+namespace ateam_common
+{
+class UDPSender
+{
+public:
+  UDPSender(
+    std::string udp_ip_address,
+    int16_t udp_port);
+
+  void send(char * data, size_t length);
```
barulicm, line 38: These should both be read-only.

```suggestion
  void send(const char * const data, const size_t length);
```

ateam_common/include/ateam_common/udp_sender.hpp
```diff
@@ -0,0 +1,54 @@
...
+#include <string>
+
+namespace ateam_common
+{
+class UDPSender
+{
+public:
+  UDPSender(
+    std::string udp_ip_address,
+    int16_t udp_port);
```
barulicm, line 36: These should both be read-only.

```suggestion
  UDPSender(
    const std::string& udp_ip_address,
    const int16_t udp_port);
```

ateam_common/src/udp_sender.cpp
```diff
@@ -0,0 +1,85 @@
...
+  io_service_.stop();
+  if (io_service_thread_.joinable()) {
+    io_service_thread_.join();
+  }
+}
+
+void UDPSender::send(char * data, size_t length)
+{
+  if (length >= buffer_.size()) {
+    // RCLCPP_ERROR(get_logger(), "UDP send data length is larger than buffer");
```
barulicm, line 58: We probably should at least log a warning if this gets tripped.

ateam_common/src/udp_sender.cpp
```diff
@@ -0,0 +1,85 @@
...
+      &UDPSender::HandleUDPSendTo, this,
+      boost::asio::placeholders::error, boost::asio::placeholders::bytes_transferred));
+}
+
+void UDPSender::HandleUDPSendTo(
+  const boost::system::error_code & error,
+  std::size_t /** bytes_transferred **/)
+{
+  if (error) {
+    // RCLCPP_ERROR(get_logger(), "Error during udp send");
```
barulicm, line 81: Same here. At least log a warning that something unexpected happened.

barulicm (commented):
I like it. :+1: 

I have a feeling we're going to end up with a utility library soon that makes publishing/subscribing to collections of topics based on robot id a little easier so not every node has to loop through the IDs and manage it itself.

ateam_ssl_simulation_radio_bridge/src/ssl_simulation_radio_bridge_node.cpp
```diff
@@ -0,0 +1,73 @@
...
+class SSLSimulationRadioBridgeNode : public rclcpp::Node
+{
+public:
+  explicit SSLSimulationRadioBridgeNode(const rclcpp::NodeOptions & options)
+  : rclcpp::Node("ateam_ssl_simulation_radio_bridge", options),
+    udp_sender_("127.0.0.1", 10301)
+  {
+    for (int robot_id = 0; robot_id < 16; robot_id++) {
+      // Copy of a callback for each robot_id
+      auto callback =
```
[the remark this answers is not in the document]
JNeiger: Super reasonable and makes the code much cleaner compared to the copy in the lambda

ateam_common/src/udp_sender.cpp
JNeiger: So I meant to ask about this, I was having problems getting this to actually compile. Is there something special I need on top of the roscpp header to be able to do Ros logs?

JNeiger: Just tried again to get the details. It seems `get_logger()` requires a ROS2 node reference (or to be called within a class that inherits from ROS2 nodes). Since this is a vanilla C++ class, I don't see an easy way to get that without just passing down a node reference :/

barulicm: That's a good point, and printing is fine for now.
Another option for situations like this is for these kinds of helpers to pass error information back up to the caller so they can decide what to do with it.

barulicm (approved):

Issue #16, opened 2021-08-11
Implement simple vision filter to estimate pos/vel of ball/robots


JNeiger:
ssl-vision gives a list of measurements per camera frame per camera for each object. These must be taken as a whole and filtered down to a single position/velocity estimate for the ball and robots.

JNeiger:
Closed from #40

PR #39, opened 2021-09-30
Vision filter using MHT with IMM KF 

JNeiger:
Halfway following [this](https://arxiv.org/pdf/1912.00603.pdf) paper about doing an MHT with tracks being able to have multiple models representing different dynamics "states". The idea is that at the very least the ball will have rolling or sliding friction which we can swap between. Hopefully, we can have multiple other models representing kicks at different speeds to more quickly react to that. I was also thinking of having a model represent just a simple bounce off a robot. With the IMM, we need some probability transition matrix representing how likely it is to move from one mode to another. This can be a function in time such that we can increase probability of kick modes when the ball is near robots.

Each camera will have a MHT for each object (ball + 32 robots). The best of each will get merged using something from [this paper](https://ieeexplore.ieee.org/stamp/stamp.jsp?tp=&arnumber=1263228) describing how good the tracks are for each camera for that object. The merged best tracks will then be output at some interval.

Closing #16

### DIFF
--- a/ateam_vision_filter/CMakeLists.txt
+++ b/ateam_vision_filter/CMakeLists.txt
@@ -1,0 +1,40 @@
+cmake_minimum_required(VERSION 3.5)
+project(ateam_ssl_vision_bridge)
+
+if(CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID MATCHES "Clang")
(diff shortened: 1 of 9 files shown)


JNeiger:
Messed up branch naming conventions. Closing and reopening with the correct convention

PR #40, opened 2021-10-03
Vision filter using MHT with IMM KF

JNeiger:
Halfway following [this](https://arxiv.org/pdf/1912.00603.pdf) paper about doing an MHT with tracks being able to have multiple models representing different dynamics "states". The idea is that at the very least the ball will have rolling or sliding friction which we can swap between. Hopefully, we can have multiple other models representing kicks at different speeds to more quickly react to that. I was also thinking of having a model represent just a simple bounce off a robot. With the IMM, we need some probability transition matrix representing how likely it is to move from one mode to another. This can be a function in time such that we can increase probability of kick modes when the ball is near robots.

Each camera will have a MHT for each object (ball + 32 robots). The best of each will get merged using something from [this paper](https://ieeexplore.ieee.org/stamp/stamp.jsp?tp=&arnumber=1263228) describing how good the tracks are for each camera for that object. The merged best tracks will then be output at some interval.

Closing #16


ateam_vision_filter/src/world.cpp
```diff
@@ -0,0 +1,206 @@
...
+        output_robot.omega += score * robot.omega;
+        output_robot.acceleration += score * robot.acceleration;
+        output_robot.alpha += score * robot.alpha;
+
+        total_score += score;
+      }
+
+      output_robot.position /= total_score;
+      output_robot.theta = 0.0;
+      if (output_angle.norm() > 0) {
```
barulicm, line 134: There's an [isZero()](https://eigen.tuxfamily.org/dox/classEigen_1_1DenseBase.html#af36014ec300f53a65083057ed4e89822) function.
```suggestion
      if (!output_angle.isZero()) {
```

ateam_vision_filter/src/world.cpp
```diff
@@ -0,0 +1,206 @@
...
+        output_robot.theta = std::atan2(output_angle.x(), output_angle.y());
```
barulicm, line 135: [std::atan2](https://en.cppreference.com/w/cpp/numeric/math/atan2) takes y first. Is the reversal of arguments intentional here?

ateam_vision_filter/test/filters/interacting_multiple_model_filter_test.cpp
```diff
@@ -0,0 +1,153 @@
...
+#include <gtest/gtest.h>
+
+#include <memory>
+#include <vector>
+
+#include "../src/filters/kalman_filter.hpp"
+#include "../src/filters/interacting_multiple_model_filter.hpp"
+#include "../src/generators/model_input_generator.hpp"
+#include "../src/generators/transmission_probability_generator.hpp"
+#include "../src/types/models.hpp"
```
barulicm, line 30: The CMake setup has `src` added as an include directory, so I think all of these test includes can remove the "../src/" prefix?

```suggestion
#include "filters/kalman_filter.hpp"
#include "filters/interacting_multiple_model_filter.hpp"
#include "generators/model_input_generator.hpp"
#include "generators/transmission_probability_generator.hpp"
#include "types/models.hpp"
```

ateam_vision_filter/test/filters/kalman_filter_test.cpp
```diff
@@ -0,0 +1,376 @@
...
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+#include "../src/filters/kalman_filter.hpp"
+
+#include <gtest/gtest.h>
+
+#include <vector>
+
+TEST(kalman_filter, getXHat_ShouldReturnIntial_WhenNoPredictOrUpdate)
```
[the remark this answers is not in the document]
JNeiger: I'm not super attached to this naming scheme, I essentially saw it on reddit a few months ago and liked the format.

If the object names are camel case, and we always do `TEST(ObjectName, testname)`, we should be all good from a technical point of view. For this specific test, we'd have `TEST(KalmanFilter, getXHat_ShouldReturnInitial_WhenNoPredictOrUpdate)`.

ateam_vision_filter/src/camera.cpp
```diff
@@ -0,0 +1,182 @@
...
+{
+  std::vector<Eigen::VectorXd> vectored_measurements;
+  vectored_measurements.reserve(robot_measurements.size());
+
+  std::transform(
+    robot_measurements.begin(), robot_measurements.end(),
+    vectored_measurements.begin(),
+    [](const RobotMeasurement & original) {
+      return Eigen::Vector3d{original.position.x(), original.position.y(), original.theta};
+    });
```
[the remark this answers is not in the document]
JNeiger: Good spot!

ateam_vision_filter/test/filters/kalman_filter_test.cpp
```diff
@@ -0,0 +1,376 @@
...
+  Eigen::Vector2d potential_measurement{11.0, 2.0};
+  Eigen::VectorXd measurement_error = kf.get_potential_measurement_error(potential_measurement);
+
+  EXPECT_NEAR(measurement_error.x(), (potential_measurement - x_hat_init).x(), 1e-4);
+  EXPECT_NEAR(measurement_error.y(), (potential_measurement - x_hat_init).y(), 1e-4);
+}
+
+TEST(kalman_filter, numeric_example)
+{
+  // Taken from here https://www.kalmanfilter.net/multiExamples.html
```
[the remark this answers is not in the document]
JNeiger: It's such a useful source to get the numbers lmao

ateam_vision_filter/src/filters/multiple_hypothesis_tracker.cpp
```diff
@@ -0,0 +1,226 @@
...
+  // Note: residual capacity so 1 is not used, 0 is used
+  edge_residual_capacity_list_t edge_residual_capacity_list = boost::get(
+    boost::edge_residual_capacity, graph);
+  for (const auto & forward_edge_vertex_pair : forward_edge_to_vertex_pair) {
+    unsigned int used_flow = 1 - edge_residual_capacity_list[forward_edge_vertex_pair.first];
+
+    // Assigned edge
+    if (used_flow == 1) {
+      auto & source_measurement_vertex = forward_edge_vertex_pair.second.first;
+      auto & sink_track_vertex = forward_edge_vertex_pair.second.second;
```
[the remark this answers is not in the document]
JNeiger: That's slick as hell

ateam_vision_filter/src/filters/kalman_filter.hpp
```diff
@@ -0,0 +1,59 @@
...
+
+private:
+  Eigen::MatrixXd F;
+  Eigen::MatrixXd B;
+  Eigen::MatrixXd H;
+  Eigen::MatrixXd Q;
+  Eigen::MatrixXd R;
+
+  Eigen::VectorXd x_hat;
+  Eigen::MatrixXd P;
```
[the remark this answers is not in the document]
JNeiger: I like that perspective as well.

I was hitting some annoyances with how to set this up programmatically without copy-pasting tons of code or over engineering it. Essentially, each camera has a list of MHT for each object (32 robot MHT's, and 1 ball MHT). In each MHT, each track is a IMM. Each model in the IMM is represented by a different KF. (This isn't super clear rereading it, I can draw a picture if it doesn't make sense).

The only difference between a robot MHT and a ball MHT is the base state space model, the list of model types being used, the input generator, and the probability generator.

The proper way to do it all would be to do dependency injection, but that feels a little too over complicate since there will only ever be two options. Having if statements everywhere isn't the most clear either. I half-assed the dependency injection by combining the input and probability generators, making everything work on Eigen::VectorXd instead of specific sizes, and passing in a specific kalman filter instantiation that will be cloned whenever we need the kalman filter.

Thoughts? It's setup in a way that it will be easy to transition in almost any direction.

ateam_vision_filter/src/camera.cpp
```diff
@@ -0,0 +1,182 @@
...
+  for (auto & yellow_robot : yellow_team) {
+    yellow_robot.set_base_track(base_track);
+  }
+
+  for (auto & blue_robot : blue_team) {
+    blue_robot.set_base_track(base_track);
+  }
+}
+
+std::vector<Eigen::VectorXd> Camera::robot_measurements_to_vector(
```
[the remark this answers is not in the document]
JNeiger: The challenge is that the MHT takes in `vector<Eigen::VectorXd>`. The compilers doesn't like converting between a `vector<Eigen::Vector3d>` and `vector<Eigen::VectorXd>`. Honestly, if you have a good suggestion for this, I hit this a lot at work and would love something more than just creating a new object with the right type and copying

ateam_vision_filter/src/filters/multiple_hypothesis_tracker.cpp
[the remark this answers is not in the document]
JNeiger: Due to it being forward_edge_vertex_pair.second.first/second, I still have an upack line in the if statement, but it's much clearer now

JNeiger:
Gotta do some end to end testing at when I get back home, but other than that and some code cleanup, this should be good to go

ateam_vision_filter/src/filters/interacting_multiple_model_filter.hpp
```diff
@@ -0,0 +1,125 @@
...
+    Eigen::MatrixXd sigma);
+
+  std::vector<Models::ModelType> model_types;  // List of models
+  std::map<Models::ModelType, KalmanFilter> models;  // Kalman filter representing ModelType
+  std::map<Models::ModelType, double> mu;  // ~= Probability of being in model ModelType
+
+  double alpha = 0.5;  // How much to weight old updates when getting average frequency
+  double relative_update_frequency = 0.5;
+  unsigned int updates_until_valid_track = 10;
+  double regularly_updated_frequncy_cutoff = 0.1;
```
JNeiger, line 119: How should we do params outside the ROS nodes? Is there an easy way or do we need to come up with some custom solution?

ateam_vision_filter/src/generators/model_input_generator.cpp
```diff
@@ -0,0 +1,248 @@
...
+    Eigen::Vector2d ball_pos = possible_state.block(0, 0, 2, 1);
+
+    std::optional<Robot> closest_robot = get_closest_robot(ball_pos);
+
+    // No visible robots
+    if (!closest_robot.has_value()) {
+      return 0.0 * possible_state;  // Easy way to get 0 vector of right size
+    }
+
+    // TODO(jneiger): Figure out bounce (assuming robot is circle for now)
```
JNeiger, line 89: Note to self, add at least a simple version of this in

ateam_vision_filter/src/generators/transmission_probability_generator.cpp
```diff
@@ -0,0 +1,190 @@
...
+  const std::optional<Robot> & robot) const
+{
+  if (!robot.has_value()) {
+    return false;
+  }
+
+  // https://en.wikipedia.org/wiki/Vector_projection
+  Eigen::Vector2d heading{cos(robot.value().theta), sin(robot.value().theta)};
+  Eigen::Vector2d robot_to_ball = position - robot.value().position;
+  double mouth_half_angle = 45.0 * 3.14 / 180.0;  // Angle between theta and side of mouth
```
JNeiger, line 175: Pull out config to the model class

ateam_vision_filter/src/message_conversions.cpp
```diff
@@ -0,0 +1,94 @@
...
+
+  return cameraFrame;
+}
+
+RobotMeasurement fromMsg(const ssl_league_msgs::msg::VisionDetectionRobot & ros_msg)
+{
+  RobotMeasurement robotDetection;
+  robotDetection.position.x() = ros_msg.pose.position.x;
+  robotDetection.position.y() = ros_msg.pose.position.y;
+  //  robotDetection.theta = ros_msg.pose.orientation;  // TODO Figure this out
```
JNeiger, line 82: Note to self, use the tf2 quat transforms

ateam_vision_filter/src/world.cpp
```diff
@@ -0,0 +1,222 @@
...
+        output_robot.acceleration += score * robot.acceleration;
+        output_robot.alpha += score * robot.alpha;
+
+        total_score += score;
+      }
+
+      output_robot.position /= total_score;
+      output_robot.theta = 0.0;
+      if (output_angle.norm() > 0) {
+        std::cout << output_angle << std::endl;
```
JNeiger, line 211:
```suggestion
```

JNeiger:
I'm out of town so I can't do the final test to make sure the refactor didn't break anything, but this should be good to review. I've been telling a robot to move around the field using the sim robot control piece, and viewing the output state info from the pub sub topic.

For now this should be good enough to make some progress. The params haven't been tuned much at all so there is quite a bit there as possible improvements. Additionally, new tracks only use the track position, but not the current best estimate velocity + accel. Should be a straight forward change in the future if it causes issue.

JNeiger:
Test looks pretty solid. I removed the some of the extra models since trying to tune everything was a little difficult just using topic echos and stuff. Each IMM is just essentially a kalman filter now.

All good to merge

JNeiger:
Gonna merge this in so `main` is basically fully featured out of the box. Very much open to fixing things up if there are any comments later.

PR #41, opened 2021-10-08
Imports ament style config for uncrustify.

barulicm:
This adds the Uncrustify config file used by the ament style linter.

With this in the repo, editors that support Uncrustify can be pointed at this file to use the standard ROS style rules.

In vs code, you can use the [Uncrustify extension](https://marketplace.visualstudio.com/items?itemName=zachflower.uncrustify). In the workspace extension settings for the Uncrustify extension, set the config path to `src/ateam_software/ament_code_style.cfg`. You can then use the "Uncrustify" formatter in vs code to format against this style (either on demand or by configuring it as the default C++ formatter for the workspace). 

image

### DIFF
--- a/ament_code_style.cfg
+++ b/ament_code_style.cfg
@@ -1,0 +1,2308 @@
+# Uncrustify-0.68
+
+#
+# General options
+#
+
+# The type of line endings. Default=Auto.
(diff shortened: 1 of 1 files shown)


JNeiger (approved):
🙌🙌

PR #43, opened 2021-10-16
Adds protobuf log handler macro.

barulicm:
While working more on the ref bridge, I ran into some protobuf errors and figured this might be useful.

By default, protobuf just logs error details to the standard error stream. The macro added here, `SET_ROS_PROTOBUF_LOG_HANDLER`, redirects all protobuf logs through the ROS logging system. Puts all of our logs in the same place for nodes that use protobuf.

It's a macro because protobuf's `SetLogHandler` only accepts a raw function pointer. Lambdas with captures can't decompose to a function pointer, so I'm using the macro to get the logger name into the lambda at compile time. Very open to ideas if anyone can figure out a cleaner way to do this.


JNeiger (approved):
I hit this same problem at work and couldn't figure a better way, nice work!

PR #45, opened 2021-11-24
ssl_league_msgs updates

barulicm:
- Adds the remaining Game Controller multicast messages
- Reorganizes messages into game_controller / vision folders
- Disables linting
  The linters run on the generated C++ code and basically just complain that the generated functions are too long. This is a result of large messages, which we're not going to change because they match the league protobufs.


ssl_league_msgs/msg/game_controller/game_events/TooManyRobots.msg
```diff
@@ -0,0 +1,4 @@
+ssl_league_msgs/Team by_team
+int32[<=1] num_robots_allowed
+int32[<=1] num_robots_on_field
+geometry_msgs/Point32[<=1] ball_location
```
JNeiger, line 4: What does the `[<=1]` mean in this context? Is it an array of length 1 or fewer similar to `optional` in proto?

JNeiger (approved):

ssl_league_msgs/msg/game_controller/game_events/TooManyRobots.msg
barulicm: Exactly. It's an admittedly pretty kludgy way to imitate optional fields.

JNeiger: Makes sense, I'm down for that

PR #46, opened 2021-11-29
Finishes Game Controller bridge nodes

barulicm:
This finishes the bridging portion of our integration with the SSL game controller. This PR includes a few high-level changes:

- Renames the ateam_game_controller_bridge package to better match names used by the league (autorefs are technically a different part of the system)
- Renames and completes gc_multicast_bridge_node. The "completion" part of this mostly involved parameterizing connection details and completing the message converters.
- Adds a new team client node that maps Game Controller team requests (ie. setting goalie ID) to ROS services.
- Gets the Game Controller working from our bringup launch file with the docker image distributed by the league. I'm not a Docker expert, so I don't know how kosher `--net host` is for something like this, but it got everything working. (Apparently, publishing multicast ports from a Docker image is tricky and not super well supported).


JNeiger (approved):
Looks good!

PR #47, opened 2021-12-08
Add joystick control

barulicm:
Adds the ateam_joystick_control package to enable joystick control of the robots. The included config file should work with any XBox-style controller and we can add others to support other types of controller. The robot being controlled is set via the node's `robot_id` parameter (which responds to runtime updates).

Along the way, I tweaked the simulator bringup file to include the bridge node and add the `headless` arg.


JNeiger (approved):
Looks good! I didn't realize ROS2 supported actually doing node level tests, that's pretty slick

PR #48, opened 2022-01-27
AI Prototype

JNeiger:
This is just a high level prototype.

There are 3 main modules. The evaluator, realization, and executors (names in flux). 

The evaluator figures out which behaviors to choose. Should we do a pass, pass then shot? Should we just shoot? What should the other robots be doing?

The realization takes a list of behaviors (or potential behaviors), does the role assignment, trajectory planning, and timing requirements. For a pass, how long will it take for the kicker to get to the ball and kick, how long will it take for the receiver to get to the receiver spot and actually complete the receive motions.

The executors takes a list of behaviors to execute, grabs the true trajectories from the realization, chooses when to  start each behavior, and does the actual trajectory following.

## High level block diagram
image

## Behavior graph example
image


ateam_ai/src/behavior.hpp
```diff
@@ -0,0 +1,36 @@
...
+    OneTouchReceiveKick,
+
+    // Kick + instantaneous target selection near kick time
+    Shot,
+    OneTouchShot,
+
+    // Normal moves
+    MoveToPoint,
+    CostFunctionPoint
+  } type;
```
JNeiger, line 19: Each of these are just direct aliases to different path planners.

ateam_ai/src/behavior.hpp
```diff
@@ -0,0 +1,36 @@
...
+
+  enum Priority {
+    Required,
+    Medium,
+    Low
+  } priority;
```
JNeiger, line 25: Required is absolutely needed for the behavior sequence, Medium and low priority describe which background behaviors to fill first if there aren't enough robots to fill them all.

JNeiger:
I think this is good to merge. It doesn't actually work yet, but it should provide a framework to add more.

ateam_ai/src/types/robot.hpp
```diff
@@ -0,0 +1,35 @@
...
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+
+#ifndef TYPES__ROBOT_HPP_
+#define TYPES__ROBOT_HPP_
+
+#include <Eigen/Dense>
+
+struct Robot
```
JNeiger, line 27: These should probably be combined with the vision filter types and moved to common.

ateam_ai/src/types/world.hpp
```diff
@@ -0,0 +1,43 @@
...
+#include "types/field.hpp"
+#include "types/referee_info.hpp"
+#include "types/robot.hpp"
+
+struct World
+{
+  Field field;
+  RefereeInfo referee_info;
+
+  std::optional<Ball> ball;
```
kylestach, line 38: Minor note, but IMO the ball should be a vector of balls and there should be a `get_unique_ball()` accessor that returns nullopt if there are zero or 2+ balls.

ateam_ai/src/behavior/behavior_feedback.hpp
```diff
@@ -0,0 +1,41 @@
...
+#include <Eigen/Dense>
+
+struct Trajectory
+{
+  // TODO(jneiger): Flesh out the trajectory structure more based on Kyle's requirements
+  Eigen::Vector2d target_point;
+};
+struct DribblerAndKickerBehavior {};
+
+struct BehaviorFeedback
```
kylestach, line 33: It seems to me that this should be different depending on the command type. Maybe a variant?

ateam_ai/src/behavior/behavior_evaluator.cpp
```diff
@@ -0,0 +1,119 @@
...
+    Behavior::Priority::Low,
+    MoveParam({10, 0})};  // Field::TheirGoal.center();
+  Behavior left_defender{
+    Behavior::Type::MoveToPoint,
+    Behavior::Priority::Medium,
+    MoveParam({5, 5})};
+  Behavior right_defender{
+    Behavior::Type::MoveToPoint,
+    Behavior::Priority::Medium,
+    MoveParam({5, -5})};
```
kylestach, line 111: I think this brings up the question of whether these should be included in the realize_behaviors call. IMO they should, but that obviously would duplicate the respective planning calls.

Maybe it makes sense to have a "coarse" and "fine" level of behavior generation? The former could just bubble up some course stat estimates while the latter does full planning?

ateam_ai/src/behavior/behavior.hpp
```diff
@@ -0,0 +1,105 @@
...
+{
+  Eigen::Vector2d target_location;
+
+  explicit MoveParam(Eigen::Vector2d target_location)
+  : target_location(target_location) {}
+};
+
+struct CostParam
+{
+  // std::function<Eigen::Vector2d()>
```
kylestach, line 65: Would be nice to also have an optional gradient function.

ateam_ai/src/ateam_ai_node.cpp
```diff
@@ -0,0 +1,62 @@
...
+
+class ATeamAINode : public rclcpp::Node
+{
+public:
+  explicit ATeamAINode(const rclcpp::NodeOptions & options)
+  : rclcpp::Node("ateam_ai", options), evaluator(realization), executor(realization)
+  {
+    DirectedGraph<Behavior> current_behaviors;
+
+    realization.update_world(world);
```
kylestach, line 43: Why not pass world into `get_best_behaviors()`/`execute_behaviors()`? I don't see a reason for it to be stateful.

ateam_ai/src/util/directed_graph.hpp
```diff
@@ -0,0 +1,82 @@
...
+// THE SOFTWARE.
+
+#ifndef UTIL__DIRECTED_GRAPH_HPP_
+#define UTIL__DIRECTED_GRAPH_HPP_
+
+#include <map>
+#include <vector>
+
+/**
+ * Directed graph of nodes with no restriction on connectivity
```
kylestach, line 28: Nit: as written this is a DAG, so there is a restriction on connectivity. I assume that's intentional because we really do not want loops in dependency graphs, but I'd mention it in the comment (unless we want to allow for the ability to modify the set of parents of a node after it's been created, but I'd prefer we _don't_ add that functionality for exactly this reason).

ateam_ai/src/util/directed_graph.hpp
```diff
@@ -0,0 +1,82 @@
...
+
+  std::vector<std::size_t> get_children(std::size_t node_idx) const
+  {
+    return parent_to_child_relationship.at(node_idx);
+  }
+
+private:
+  std::vector<Node> nodes;
+  std::vector<std::size_t> root_nodes;
+  std::map<std::size_t, std::vector<std::size_t>> parent_to_child_relationship;
```
kylestach, line 79: Nit: prefer `unordered_map`.

ateam_ai/src/behavior/behavior_feedback.hpp
JNeiger: That's a good thought. I'll swap it over so it's more easily extendable as we have more visibility into the feedback requirements.

ateam_ai/src/ateam_ai_node.cpp
JNeiger: It doesn't need to be stateful. I'll swap it

ateam_ai/src/behavior/behavior_evaluator.cpp
JNeiger: I like the idea of a coarse and fine version. On the high level side, I don't really care about how the behavior happens, but the results of it.

I was splitting these behaviors out from the above set because I was kinda thinking about it in the "primary"/"passive" role split. I really only care about the main ball interaction behaviors (Eg: on offense the capture/pass/shot). The background defenders don't change my decision making in most cases so it would be a waste to calculate them. If they do matter, we can add them to `realize_behaviors` call and get the feedback we need.

I can add them to the original call for now so we don't have any extra requirements on the role assignment piece. It might be something to think about in the future since it encodes background "formations" (for lack of a better word) in a more obvious way. We won't have to plan out all 11 robot's behaviors when trying do decide which plan to do.

kylestach: Maybe we could have an explicit `PassiveFormation` behavior or something? IMO we should guarantee that all behaviors are included in the evaluation, even if we aren't coming up with full plans for all of them.

JNeiger:
Gonna abuse my powers and merge this in. :P

PR #49, opened 2022-01-31
Sim robot feedback

JNeiger:
Finishes out the part I forgot in #2 

Provides ball sense to enter ros world

### DIFF
--- a/ateam_ssl_simulation_radio_bridge/src/ssl_simulation_radio_bridge_node.cpp
+++ b/ateam_ssl_simulation_radio_bridge/src/ssl_simulation_radio_bridge_node.cpp
@@ -21,9 +21,11 @@
 #include <rclcpp/rclcpp.hpp>
 #include <rclcpp_components/register_node_macro.hpp>
 
-#include <ateam_common/udp_sender.hpp>
+#include <ateam_common/bi_directional_udp.hpp>
+#include <ateam_msgs/msg/robot_feedback.hpp>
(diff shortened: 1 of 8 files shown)


ateam_ssl_simulation_radio_bridge/src/ssl_simulation_radio_bridge_node.cpp
```diff
@@ -67,14 +76,28 @@ class SSLSimulationRadioBridgeNode : public rclcpp::Node
...
     if (robots_control.SerializeToString(&protobuf_msg)) {
-      udp_sender_.send(protobuf_msg.data(), protobuf_msg.size());
+      udp_.send(protobuf_msg.data(), protobuf_msg.size());
+    }
+  }
+
+  void feedback_callback(const char * buffer, size_t bytes_received)
+  {
+    RobotControlResponse feedback_proto;
+    if (!feedback_proto.ParseFromArray(buffer, bytes_received - 1)) {
```
JNeiger, line 86: Same issue as the vision filter. It really does not like reading the entire packet. Someone else should probably try this since it may be a WSL issue

chachmu (approved):
The code looks good although the packet thing still kind of weirds me out. I did check and whenever I publish a command to a robot I get back a corresponding feedback message so clearly it isn't segfaulting or anything and the message made sense so its probably fine 🤷‍♂️

PR #50, opened 2022-02-13
UI Build Changes

barulicm:
I ended up just splitting up the 'dependency installation' and 'build' steps into different scripts, to avoid trying to run `sudo` from CMake. These changes build the javascript stuff in the source directory and then install it, which should play nicely with symlink installs for hot reload things.

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,6 +23,14 @@
    rosdep install --from-paths . --ignore-src -y
    ```
 
+1. Install our non-ROS dependencies
+
+   ```bash
+   # In the ateam_ws directory
+   source /opt/ros/galactic/setup.bash
+   ./ateam_ui/install_deps.sh
+   ```
+
 1. Build the code
 
    ```bash
--- a/README.md
+++ b/README.md
@@ -28,7 +28,7 @@
    ```bash
    # In the ateam_ws directory
    source /opt/ros/galactic/setup.bash
-   ./src/ateam_ui/install_deps.sh
+   ./src/software/ateam_ui/install_deps.sh
    ```
 
 1. Build the code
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -23,6 +23,14 @@
    rosdep install --from-paths . --ignore-src -y
    ```
 
+1. Install our non-ROS dependencies
+
+   ```bash
+   # In the ateam_ws directory
+   source /opt/ros/galactic/setup.bash
+   ./ateam_ui/install_deps.sh
```
chachmu, line 31: Unless we have our workspaces set up differently the path for me was `./src/software/ateam_ui/install_deps.sh`. Also, when I tried to run it on a fresh installation it just says "Upgrading node to latest stable" every time and then doesn't actually install any of the packages other than npm.

chachmu:
I think your if statement in the install_deps.sh might be wrong. The npm list command returns a non-zero exit status to indicate that it didn't find the package but the if statement is testing for success. It seems like it works if you replace line 11 with this:
`if !npm list --depth 1 --global $package_name > /dev/null 2>&1; then`

barulicm:
Yeah, I just figured that out too. Testing things in a local vm right now.
My bash is bad and I should feel bad.

barulicm:
Alright, turns out installing that webkit package requires installing a _buuunch_ of other GUI / front-end related packages.
My personal vote would be to skip the UI tests in CI for now (as is implemented in this PR now), and we can spend some time poking at this in the background and see if there's maybe a different base image we can use or something to get UI tests running in CI again.

As things stand in this PR right now, CI works at least to tell us if the UI build works.
https://github.com/SSL-A-Team/software/runs/5256410208

Issue #54, opened 2022-05-02
Allow control of more than 1 robot


JNeiger:
Only 1 robot is able to be controlled at once in the behavior executor

Issue #53, opened 2022-05-02
Fix UI to scale with default grSim field


JNeiger:
I think it's 4x smaller than it should be, so it might just be a div A/B problem.

PR #55, opened 2022-05-02
Integrate AI into the rest of the system

### LINKED ISSUE #19: Add basic AI
Something as simple as commanding a robot to move in a straight line from point A to point B so we can get some feedback into the rest of the system during tests.

JNeiger:
- Integrates a simple move to ball command for our team
- Adds piping for vision to show up on the UI

Closes #19 

Add #52 #53 #54 as future TODO's to keep things on track


ateam_ai/src/behavior/behavior_executor.cpp
```diff
@@ -22,8 +22,10 @@
 
 #include "behavior/behavior_feedback.hpp"
 
-BehaviorExecutor::BehaviorExecutor(BehaviorRealization & behavior_realization)
-: behavior_realization(behavior_realization) {}
+BehaviorExecutor::BehaviorExecutor(
+  BehaviorRealization & behavior_realization,
+  rclcpp::Publisher<ateam_msgs::msg::RobotMotionCommand>::SharedPtr robot_commands)
+: behavior_realization(behavior_realization), robot_commands(robot_commands) {}
```
JNeiger, line 28: I'm don't super liking trying to pass down the motion command pubs, I think I might change this to return an array of motion commands (or even just return a list of ros msgs), and have a separate module that converts motion commands to the ros msg. I'll do that in another PR to clean this up though

ateam_ui/src/src/components/FieldComponent.vue
```diff
@@ -145,8 +145,8 @@
                 dragBoundFunc: dragBound,
                 r_id: robot.id,
                 team: robot.team,
-                x: robot.pose.position.x,
-                y: robot.pose.position.y,
+                x: robot.pose.position.y / 4,
```
chachmu, line 148: I don't think this is the best place to modify how the robots are positioned. What specifically are you trying to get the UI to do?

chachmu: If the robots are showing up in the wrong position on the UI then I think I need to set up the UI to also pull the field dimensions from the vision packet so that the scaling on the field and the robots match.

JNeiger: It just had the wrong scaling. I'll revert this change for now and we can add that feature in later

JNeiger:
Gonna merge this

PR #56, opened 2022-06-10
Allow for control of all robots

### LINKED ISSUE #54: Allow control of more than 1 robot

JNeiger:
Closes #54

Pulled the publisher out of the behavior executor.

### DIFF
--- a/ateam_ai/src/behavior/behavior_executor.cpp
+++ b/ateam_ai/src/behavior/behavior_executor.cpp
@@ -22,12 +22,10 @@
 
 #include "behavior/behavior_feedback.hpp"
 
-BehaviorExecutor::BehaviorExecutor(
-  BehaviorRealization & behavior_realization,
-  rclcpp::Publisher<ateam_msgs::msg::RobotMotionCommand>::SharedPtr robot_commands)
-: behavior_realization(behavior_realization), robot_commands(robot_commands) {}
+BehaviorExecutor::BehaviorExecutor(BehaviorRealization & behavior_realization)
+: behavior_realization(behavior_realization) {}
 
-void BehaviorExecutor::execute_behaviors(
+BehaviorExecutor::RobotMotionCommands BehaviorExecutor::execute_behaviors(
   const DirectedGraph<Behavior> & behaviors,
   const World & world)
 {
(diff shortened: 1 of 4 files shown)


ateam_ssl_simulation_radio_bridge/src/ssl_simulation_radio_bridge_node.cpp
```diff
@@ -58,7 +58,7 @@ class SSLSimulationRadioBridgeNode : public rclcpp::Node
 
       command_subscriptions_.at(robot_id) =
         create_subscription<ateam_msgs::msg::RobotMotionCommand>(
-        "~/robot_motion_commands/robot" + std::to_string(robot_id),
+        "/ateam_ai/robot_motion_commands/robot" + std::to_string(robot_id),
```
barulicm, line 61: I'm curious what motivated this change instead of remapping topics in the associated launch file?

ateam_ai/src/behavior/behavior_executor.cpp
```diff
@@ -50,15 +48,25 @@ void BehaviorExecutor::execute_behaviors(
...
+    // Note: This abuses the way the behaviors are supposed to be executed
+    // This is just a placeholder for now
+    Eigen::Vector2d target = std::get<MoveParam>(
+      behaviors.get_node(
+        behaviors.get_root_nodes().front()).params).target_location;
+
+    Eigen::Vector2d command = (target - world.our_robots.at(id).value_or(Robot()).pos) / 1000.0;
+
+    if (command.norm() > 1) {
+      command = command.normalized();
```
barulicm, line 62: There's a function for normalizing the vector in place.

```suggestion
      command.normalize();
```

barulicm (approved):
Approving with suggestions.

ateam_ssl_simulation_radio_bridge/src/ssl_simulation_radio_bridge_node.cpp
JNeiger: Honestly, I figured that with 16 robots, I mess up less just editing the sub topic instead of trying to copy paste that remapping 16 times. All the other pub sub relationships are hardcoded as well. I don't know the "right" way to do this in ROS, so I very open to change it

barulicm: I _believe_ you can remap base names for multiple topics in one remap rule in ROS 2, though I can't remember if I've tried that before.

There are two popular conventions for handling topic names, though there's a ton of code that doesn't follow either out there in the world, since nothing enforces it.
- Nodes default to global topic names, such as `/cmd_vel` and `/image_raw`, which are then remapped to specific names, such as `/camera/front/color/image_raw`
   Under this pattern, we'd use `/robot_motion_commands/robot0`
- Nodes default to private topic names, such as `/my_controller/cmd_vel`, which are then remapped to specific names.
   Under this pattern, we'd use `~/robot_motion_commands/robot0`

Of course, this is all a little arbitrary since everything works as long as you sort out the names in the end. I commented mostly out of curiosity if you had run into some problem with using the locally namespaced name.

JNeiger: I gotcha. In the long term I think it's going to be required that we refactor this into a cleaner convention, but I don't think it's required right this second. I'll make an issue for it so we can track the work

PR #57, opened 2022-06-10
Convert units to m from mm

### LINKED ISSUE #52: Change position to be meters not mm

JNeiger:
Closes #52

### DIFF
--- a/ateam_ssl_vision_bridge/src/message_conversions.cpp
+++ b/ateam_ssl_vision_bridge/src/message_conversions.cpp
@@ -28,7 +28,7 @@
 namespace ateam_ssl_vision_bridge::message_conversions
 {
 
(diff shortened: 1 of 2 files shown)


ateam_ssl_vision_bridge/src/message_conversions.cpp
```diff
@@ -28,14 +28,16 @@
 namespace ateam_ssl_vision_bridge::message_conversions
 {
 
+constexpr double mmTom = 1.0 / 1000;
```
barulicm, line 31: nit / alternative: You can use scientific notation in C++.
```suggestion
constexpr double mmTom = 1.0e-3;
```

barulicm (approved):
Approving with one optional suggestion.

PR #58, opened 2022-06-12
Fixes the robot ID color patterns in the UI status panel.

barulicm:
This resolves a TODO in the UI code about the robot ID color patterns being wrong. Unfortunately, they are _not_ just the binary representation of the ID number. They are an arbitrary mapping of ID to unique pattern, defined in [rule 3.2.4](https://robocup-ssl.github.io/ssl-rules/sslrules.html#_vision_pattern).


chachmu (approved):
:+1:

Issue #14, opened 2021-08-11
Interface with UI referee


JNeiger:
Since the UI will also be able to send referee commands, we must also be able to take UI ref commands and convert to the local ref object.

barulicm:
UI ref commands go through the GC, so the data path is the same between command sources.

PR #59, opened 2022-06-14
First pass at in-GUI Game Controller buttons

barulicm:
This adds an MVP GUI component that provides the 4 basic game controller command buttons (start, force start, stop, and halt). Still lots of room for making this better, but I wanted to start small and make sure I'm not way off track with the vue / JS stuff before getting too far in.

These buttons talk directly to the GC using its "API websocket." (Same way the GC's GUI sends updates.) Any updates triggered by our UI are thus indistinguishable from those triggered by the GC's GUI, so they propagate through our system via our existing bridge and ROS topics.

Related to #14 and #6

### DIFF
--- a/ateam_ui/src/src/App.vue
+++ b/ateam_ui/src/src/App.vue
@@ -5,6 +5,7 @@
         </v-app-bar>
         <v-main>
             <v-container class="d-flex flex-row" ref="Main Components">
+                <RefButtonsComponent/>
                 <StatusComponent/>
                 <FieldComponent/>
             </v-container>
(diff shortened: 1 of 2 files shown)


chachmu:
I've tested it and everything works perfectly when the ssl game controller node is running but it might be convenient to figure out a way to not need to have that node running when we are in simulation since the UI is the only way we would try to send ref commands anyway. I haven't actually started figuring out how easy that is to do though.

barulicm:
My plan was just to run the game controller node in simulation as well. Anything else would require re-implementing our own version of the game controller server, which doesn't feel worth it.
Is there something specific you want to do with sim that the normal game controller server doesn't support?

chachmu:
I talked with Joe about this in the last meeting and I think running the game controller node should be fine

chachmu (approved):

PR #61, opened 2022-06-20
Add trapezoidal motion planning

JNeiger:
Part 1: Add a trapezoidal motion planner so we can do A->B planning. I fully expect this to get replaced at a later point in time.

### DIFF
--- a/ateam_ai/src/trajectory_generation/trapezoidal_motion_profile.cpp
+++ b/ateam_ai/src/trajectory_generation/trapezoidal_motion_profile.cpp
@@ -19,15 +19,17 @@
 // THE SOFTWARE.
 
 #include "trajectory_generation/trapezoidal_motion_profile.hpp"
-#include <iostream>
+
+#include <numeric>
+
 namespace TrapezoidalMotionProfile
 {
 Trajectory Generate3d(
-  Eigen::Vector3d start, Eigen::Vector3d start_vel,
-  Eigen::Vector3d end, Eigen::Vector3d end_vel,
-  Eigen::Vector3d max_vel_limits,
-  Eigen::Vector3d max_accel_limits,
-  double dt)
+  const Eigen::Vector3d & start, const Eigen::Vector3d & start_vel,
+  const Eigen::Vector3d & end, const Eigen::Vector3d & end_vel,
+  const Eigen::Vector3d & max_vel_limits,
+  const Eigen::Vector3d & max_accel_limits,
+  const double dt)
 {
   // Independently plan for each DOF
   std::array<Trajectory1d, 3> trajectories;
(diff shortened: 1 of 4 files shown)


JNeiger:
I expect this to fail to build as I could only test with #62 included. I split the PR's up though so it should be easier to review.

barulicm (approved):
Approved with some suggestions.

ateam_ai/src/trajectory_generation/trapezoidal_motion_profile.cpp
```diff
@@ -0,0 +1,244 @@
...
+
+  // Distance to cover
+  double target_dist = end_pos - start_pos;
+
+  // Time needed to accel/decel to/from max
+  double t_start_to_max_vel = std::abs(max_vel - start_vel) / max_accel;
+  double t_max_to_end_vel = std::abs(end_vel - max_vel) / max_accel;
+
+  // Avg velocity over accel/decel
+  double avg_vel_start_to_max_vel = (start_vel + max_vel) / 2;
```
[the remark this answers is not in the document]
JNeiger: That's actually super nice, thanks for sharing!

JNeiger:
I've split up that massive function into a bunch of small functions to clean things up. Hopefully this is a lot cleaner read

ateam_ai/src/trajectory_generation/trapezoidal_motion_profile.cpp
```diff
@@ -0,0 +1,360 @@
...
+  PhaseCoeffs phase_1;
+  PhaseCoeffs phase_2;
+  PhaseCoeffs phase_3;
+};
+
+Trajectory1d sample_trapezoidal_profile(
+  const TrapezoidalCoeffs coeffs, const double accel_limit,
+  const double dt)
+{
+  const auto sq = [](const double x) {return x * x;};
```
barulicm, line 165: I wonder if we could just use `std::pow(___,2);`.

Would mostly be a style/readability question. Performance seems to be effectively equivalent (https://quick-bench.com/q/ITvfXrs_QhDiPLTJwYhu33u4uzo)

ateam_ai/src/trajectory_generation/trapezoidal_motion_profile.cpp
```diff
@@ -0,0 +1,360 @@
...
+  const auto position = [sq](const double t, const double v_initial, const double a) {
+      return v_initial * t + 1. / 2. * a * sq(t);
+    };
+  const auto velocity = [sq](const double t, const double v_initial, const double a) {
+      return v_initial + a * t;
+    };
```
barulicm, line 171: Captures, but doesn't use `sq`
```suggestion
  const auto velocity = [](const double t, const double v_initial, const double a) {
      return v_initial + a * t;
    };
```

ateam_ai/src/trajectory_generation/trapezoidal_motion_profile.cpp
```diff
@@ -0,0 +1,360 @@
...
+    trapezoidal_coeffs.phase_3.duration = phase_3_from_nonlimit_vel.time_required;
+    trapezoidal_coeffs.phase_3.end_vel = end_vel;
+  }
+
+  Trajectory1d output = sample_trapezoidal_profile(trapezoidal_coeffs, max_accel, dt);
+
+  // Add offset start pos
+  for (auto & sample : output.samples) {
+    sample.pos += start_pos;
+  }
```
barulicm, line 348: Consider [std::accumulate](https://en.cppreference.com/w/cpp/algorithm/accumulate)

```suggestion
  sample.pos = std::accumulate(output.samples.begin(), output.samples.end(), sample.pos);
```

barulicm (approved):
The refactor helped a lot for understanding!

ateam_ai/src/trajectory_generation/trapezoidal_motion_profile.cpp
JNeiger: You right, swapped it over to `pow`. The normal way is def clearer in this context.

ateam_ai/src/trajectory_generation/trapezoidal_motion_profile.cpp
JNeiger: I'm not sure `std::accumulate` is the right one, this is adding the start position offset to all samples. I think this would be the way to go if  we wanted to use `<algorithm>`
```
std::transform(output.samples.begin(),
               output.samples.end(),
               output.samples.begin(),
               [start_pus](double p) {
                 return p + start_pos;
               });
```

JNeiger: Actually, I'll just swap to that since it's not clear with just a glance

PR #65, opened 2022-06-26
Changes the UI field dimensions to match Div A rules.

### LINKED ISSUE #53: Fix UI to scale with default grSim field

barulicm:
Fixes #53 

Dimensions pulled from [figure 1 in section 2.1.1 in the league rules](https://robocup-ssl.github.io/ssl-rules/sslrules.html#field-dimensions-a).

Scaling adjust to keep new field similar in size to previous field render.

### DIFF
--- a/ateam_ui/src/src/App.vue
+++ b/ateam_ui/src/src/App.vue
@@ -47,27 +47,28 @@ export default {
                 },
                 underlays: {},
                 overlays: {},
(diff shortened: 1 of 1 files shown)


chachmu (approved):
LGTM :+1:

PR #68, opened 2022-08-09
Add immutable duration

JNeiger:
I need to clean this up and split it so it's not so many changes. Each bullet point below will hopefully be a different PR so that it's stupid easy to review

Included is
- Immutable duration so the first X seconds of the trajectory from last time stays constant
- Trajectory editing/splicing/cutting etc to support messing with trajectories + unit tests
- The idea of self state for each module to try and keep all the behavior stuff stateless
- Split behavior follower module from behavior executor to better split the trajectory generation from following


JNeiger:
Gonna keep this as the main PR to merge in since I keep forgetting how to do stacked PR's in github nicely. I'll have temp PR's for each commit, once all are approved, I'll just merge this instead

JNeiger:
All other PR's are approved, gonna merge this once CI passes

PR #70, opened 2022-08-10
Split executor into executor and follower (#2)

JNeiger:
Splits the behavior_executor into the trajectory generation and the controller piece. 

Part 2 for #68


chachmu (approved):
I fixed a typo in a comment but everything else looks good

PR #73, opened 2022-08-10
Add immutable duration (#5)

JNeiger:
Adds an immutable duration which results in the next X seconds of the trajectory being locked in and unable to be replanned. 

Adds `plan_from` robot states which are the robot states at the end of the immutable duration. All planning is done using the planned position instead of the actual measured position. It's up to the behavior_follower to keep up

Part 5 for #68

### DIFF
--- a/ateam_ai/src/ateam_ai_node.cpp
+++ b/ateam_ai/src/ateam_ai_node.cpp
@@ -26,6 +26,7 @@
 #include <tf2_geometry_msgs/tf2_geometry_msgs.h>
 #include <tf2/LinearMath/Quaternion.h>
 
+#include <array>
 #include <chrono>
 #include <functional>
 #include <mutex>
@@ -37,6 +38,7 @@
 #include "behavior/behavior_executor.hpp"
 #include "behavior/behavior_follower.hpp"
 #include "behavior/behavior_realization.hpp"
+#include "trajectory_generation/trajectory_editor.hpp"
 #include "types/world.hpp"
 #include "util/directed_graph.hpp"
 
@@ -107,6 +109,7 @@ private:
   BehaviorFollower follower_;
   std::mutex world_mutex_;
   World world_;
+  std::array<std::optional<Trajectory>, 16> previous_frame_trajectories;
 
   void robot_state_callback(
     std::array<std::optional<Robot>, 16> & robot_states,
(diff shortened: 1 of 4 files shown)


ateam_ai/src/behavior/behavior_executor.cpp
```diff
@@ -64,17 +64,17 @@ std::array<std::optional<Trajectory>, 16> BehaviorExecutor::execute_behaviors(
...
       std::size_t assigned_robot = root_node.assigned_robot_id.value();
 
-      Sample3d command;
-      if (root_node.trajectory.samples.size() >= 2) {
-        command = root_node.trajectory.samples.at(1);  // Where I want to be next frame
-      } else if (root_node.trajectory.samples.size() == 1) {
-        // Already at target since only sample is current position
-        command = root_node.trajectory.samples.front();
-      } else {
-        continue;
```
JNeiger, line 74: This removal is an artifact of git :/
Just assume this was already gone

ateam_ai/src/behavior/behavior_executor.cpp
```diff
@@ -64,17 +64,17 @@ std::array<std::optional<Trajectory>, 16> BehaviorExecutor::execute_behaviors(
...
+        const double dt = 0.01;
+        trajectory = trajectory_editor::apply_immutable_duration(
+          self_state.previous_trajectories.at(assigned_robot).value(),
+          trajectory,
+          immutable_duration,
+          dt, world.current_time);
       }
 
-      output_trajectories.at(assigned_robot) = root_node.trajectory;
+      output_trajectories.at(root_node.assigned_robot_id.value()) = root_node.trajectory;
```
JNeiger, line 77: This change also shouldn't have happened

ateam_ai/src/trajectory_generation/trajectory_generation.cpp
```diff
@@ -64,10 +67,12 @@ BehaviorFeedback TrajectoryGeneration::get_feedback_from_behavior(
         Eigen::Vector3d max_vel{2, 2, 0.5};  // TODO(jneiger): Set as params
         Eigen::Vector3d max_accel{2, 2, 0.5};
         double dt = 0.01;  // TODO(jneiger): Feed this down from above
-        feedback.trajectory = TrapezoidalMotionProfile::Generate3d(
+        Trajectory trajectory = TrapezoidalMotionProfile::Generate3d(
           current, current_vel, target,
           target_vel, max_vel, max_accel,
           dt, world.current_time);
+
+        feedback.trajectory = trajectory;
```
JNeiger, line 75: I believe this change is also accidently added due to git

ateam_ai/src/ateam_ai_node.cpp
```diff
@@ -144,6 +147,19 @@ class ATeamAINode : public rclcpp::Node
     // Preproccess world
     //
     world_.current_time += 0.01;
+    for (int i = 0; i < 16; i++) {
```
chachmu, line 150: It seems like in other places we are iterating through robot ids we use `std::size_t robot_id` as the variable which is more obvious what is happening

chachmu (approved):

PR #76, opened 2022-09-23
Add iLQR interface

JNeiger:
Adds an abstract iLQR problem interface and solver. Ideally, we'll be able to use this to generate trajectories for movement + capture + kicks.

Features
- Overridable interface for dynamics and cost functions as a function of time
- Automatic differentiation for all dynamics and cost
- Tune-able max iteration count + alpha gain to control convergence

Not Features
- Only supports dynamics with no second derivative (I didn't want to deal with tensor math lol)

### DIFF
--- a/ateam_ai/src/trajectory_generation/ilqr_problem.hpp
+++ b/ateam_ai/src/trajectory_generation/ilqr_problem.hpp
@@ -78,7 +78,7 @@ public:
       State x_eps_n = x;
       x_eps_p(i) += eps;
       x_eps_n(i) -= eps;
-      out(i) = (cost(x_eps_p, u, t) - cost(x_eps_n, u, t)) / (2*eps);
+      out(i) = (cost(x_eps_p, u, t) - cost(x_eps_n, u, t)) / (2 * eps);
     }
 
     for (std::size_t i = 0; i < U; i++) {
@@ -86,7 +86,7 @@ public:
       Input u_eps_n = u;
       u_eps_p(i) += eps;
       u_eps_n(i) -= eps;
-      out(i + X) = (cost(x, u_eps_p, t) - cost(x, u_eps_n, t)) / (2*eps);
+      out(i + X) = (cost(x, u_eps_p, t) - cost(x, u_eps_n, t)) / (2 * eps);
     }
 
     return out;
@@ -113,11 +113,8 @@ public:
 
   Trajectory calculate(const State & initial_state)
   {
-    std::cout << "Forward rollout" << std::endl;
     forward_rollout(initial_state);
-    std::cout << "Backward pass" << std::endl;
     backward_pass();
-    std::cout << "Return" << std::endl;
 
     return trajectory;
   }
@@ -148,8 +145,6 @@ private:
       g.at(t) = gradiant(trajectory.at(t), actions.at(t), t);
       h.at(t) = hessian(trajectory.at(t), actions.at(t), t, g.at(t));
     }
-
-    std::cout << "Initial cost " << overall_cost << std::endl;
   }
 
   void backward_pass()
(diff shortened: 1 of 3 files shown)


ateam_ai/src/trajectory_generation/ilqr_problem.hpp
```diff
@@ -0,0 +1,290 @@
...
+      g.at(t) = gradiant(trajectory.at(t), actions.at(t), t);
+      h.at(t) = hessian(trajectory.at(t), actions.at(t), t, g.at(t));
+    }
+  }
+
+  void backward_pass()
+  {
+    std::size_t num_iterations = 0;
+
+    while (num_iterations < max_num_iterations) {
```
barulicm, line 154: Could this be a for loop? I don't see any continues that would prevent the increment line at the end of the loop body begin run every iteration.

ateam_ai/src/trajectory_generation/ilqr_problem.hpp
```diff
@@ -0,0 +1,290 @@
...
+  using Hessians = std::array<Hessian, T>;
+  using Gradiants = std::array<Gradiant, T>;
+
+
+  void forward_rollout(const State & initial_state)
+  {
+    // Step 1: Forward Rollout
+    trajectory.front() = initial_state;
+    actions.front().setZero();
+    overall_cost = cost(trajectory.at(0), actions.at(0), 0);
```
barulicm, line 133: nit: consider `std::array<>::front()`
```suggestion
    overall_cost = cost(trajectory.front(), actions.front(), 0);
```

ateam_ai/src/trajectory_generation/ilqr_problem.hpp
```diff
@@ -0,0 +1,290 @@
...
+          j.at(t).block(0, X, X, U);
+
+        // eq 5b
+        // Solve inverse with regulization to keep it from exploding
+        Eigen::SelfAdjointEigenSolver<Eigen::Matrix<double, U, U>> eigensolver(Q_ddxu.block(
+            X, X, U,
+            U));
+        if (eigensolver.info() != Eigen::Success) {
+          std::cout << "Failed to solve eigen vectors" << std::endl;
+          abort();
```
barulicm, line 204: Consider throwing an exception instead so the node that ends up wrapping this can correctly log and maybe recover?

ateam_ai/src/trajectory_generation/ilqr_problem.hpp
```diff
@@ -0,0 +1,290 @@
...
+        // eq 6c
+        v_ddx =
+          Q_ddxu.block(0, 0, X, X) - K.at(t).transpose() * Q_ddxu.block(X, X, U, U) * K.at(t);
+      }
+
+      // Step 3.5: Update possible control signal, states, costs
+      Cost test_cost;
+      Trajectory test_trajectory;
+      Actions test_actions;
+      test_trajectory.at(0) = trajectory.at(0);
```
barulicm, line 233: nit: Couple more places to replace `at(0)` with `front()`

ateam_ai/src/trajectory_generation/ilqr_problem.hpp
JNeiger: Sure! It's just the basic run until convergence or max number of iterations so nothing too fancy :P

ateam_ai/src/trajectory_generation/ilqr_problem.hpp
JNeiger: Oops, this was just for debugging. For the short term, I might make the return a bool signaling success instead of doing exceptions.

In the long term, I would like to integrate `absl::Status` into our code base. I've really enjoyed our custom version of this at work and it makes investigating crashes super simple.

barulicm (approved):

PR #78, opened 2022-10-14
Log world to ros topic

JNeiger:
At a later point in time, we can read this topic to "run" the AI step by step. If there are any specific crashes or weird behavior, we can reproduce that exact frame quite easily.

### DIFF
--- a/ateam_msgs/msg/World.msg
+++ b/ateam_msgs/msg/World.msg
@@ -1,4 +1,4 @@
-float32 current_time
+builtin_interfaces/Time current_time
 
 # Field field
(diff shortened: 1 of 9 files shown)


ateam_msgs/msg/World.msg
```diff
@@ -0,0 +1,12 @@
+float32 current_time
```
barulicm, line 1: Should this be a `builtin_interfaces/Time` field instead?

JNeiger: True true, forgot that was a thing

PR #83, opened 2022-12-13
minor fixes from setup on a fresh OS

guyfleeman:
I recently had to reinstall 20.04 so I have a fairly pristine system. A few setup things didn't work. I've worked through some and I touched up:
- it seems necessary to source the underlay before installing our dependencies in step `3`, this is reflected in the readme
- bash compatible shells (like ZSH) don't work if sourcing setup.bash. This silently fails to set a few env vars. A reminder is added to source the correct setup file for your shell since it's silent. This applies to `most steps`.
- the out-of-ROS dependency script in ateam_ui silently suppresses errors if npm or nodejs are not installed. The script now prints the missing core dependency and attempts to install it. This applies to step `4`

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,6 +6,13 @@
 
    Follow the [instructions for ROS 2 Galactic](http://docs.ros.org/en/galactic/Installation.html) for your system.
 
+   Initialize rosdep with the following
+   ```bash
+   sudo rosdep init
+   # galactic went EOL Dec 1 2022 so we need the flag
+   rosdep update --include-eol-distros
+   ```
+
 1. Make a ROS workspace and clone our repo
 
    ```bash
@@ -17,17 +24,16 @@
    ```
 
 1. Install our ROS dependencies
-   Remember to source the underlay script corresponding to your shell (.sh, .bash, and .zsh are available). 
 
    ```bash
    # source the underlay
    source /opt/ros/galactic/setup.bash
+
    # In the ateam_ws directory
    rosdep install --from-paths . --ignore-src -y
    ```
 
 1. Install our non-ROS dependencies
-   Remember to source the underlay script corresponding to your shell (.sh, .bash, and .zsh are available). 
 
    ```bash
    # In the ateam_ws directory
@@ -37,10 +43,11 @@
 
 1. Build the code
 
+
    ```bash
    # In the ateam_ws directory
    source /opt/ros/galactic/setup.bash
    colcon build
    ```
 
-**Note:** You'll need to source both the underlay (`/opt/ros/galactic/setup.bash`) and our workspace's overlay (`ateam_ws/install/setup.bash`) in every terminal session before running any of our code.
+**Note:** You'll need to source both the underlay (`/opt/ros/galactic/setup.bash`) and our workspace's overlay (`ateam_ws/install/setup.bash`) in every terminal session before running any of our code. Remember to source the underlay script corresponding to your shell (.sh, .bash, and .zsh are available). 
--- a/README.md
+++ b/README.md
@@ -43,7 +43,6 @@
 
 1. Build the code
 
-
    ```bash
    # In the ateam_ws directory
    source /opt/ros/galactic/setup.bash
--- a/README.md
+++ b/README.md
@@ -17,13 +17,17 @@
    ```
 
 1. Install our ROS dependencies
+   Remember to source the underlay script corresponding to your shell (.sh, .bash, and .zsh are available). 
 
    ```bash
+   # source the underlay
+   source /opt/ros/galactic/setup.bash
    # In the ateam_ws directory
    rosdep install --from-paths . --ignore-src -y
    ```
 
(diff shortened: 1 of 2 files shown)


guyfleeman:
I'm still getting the following error when installing our dependencies. Any ideas what's going on?


```
ERROR: the following packages/stacks could not have their rosdep keys resolved
to system dependencies:
ateam_ui: Cannot locate rosdep definition for [rosbridge_server]
```

ateam_ui/install_deps.sh
```diff
@@ -6,10 +6,22 @@
...
+    echo "failed to install npm"
+    exit 1
+  fi
+fi
+
 install_npm_package_if_missing () {
   local package_name=$1
   echo "Checking for npm package: $package_name"
-  if npm list --depth 1 --global $package_name > /dev/null 2>&1; then
+  if npm list --depth 1 --global $package_name > /dev/null; then
```
guyfleeman, line 24: don't suppress stderr if we get here. `set -e` is flagged above so the script will quit

README.md
```diff
@@ -17,13 +17,17 @@
    ```
 
 1. Install our ROS dependencies
+   Remember to source the underlay script corresponding to your shell (.sh, .bash, and .zsh are available).
```
barulicm, line 20: Need a blank line between this and the numbered step to actually get rendered on a new line.

ateam_ui/install_deps.sh
barulicm: I'm curious if you hit a specific problem that motivated this change?
This command call is part of a conditional, so `set -e` should ignore it for the sake of its error checking. We're in fact counting on that behavior here, since this command returning 1 is our indicator that the package given needs to be installed.
I don't necessarily disagree with the change. I'm just not following the explanation.

ateam_ui/install_deps.sh
```diff
@@ -6,10 +6,22 @@
...
+  echo "ERROR: npm is not installed"
+
+  echo "attempting to install nodejs and npm"
+  sudo apt update -y
+  sudo apt install nodejs npm -y
+  if ! command -v npm; then
+    echo "failed to install npm"
+    exit 1
+  fi
+fi
```
barulicm, line 19: npm should have been installed by rosdep (because it's listed as a dependency in the package manifest for ateam_ui)
nodejs is a dependency of npm, so it also should've been installed in the rosdep step.

barulicm:
> I'm still getting the following error when installing our dependencies. Any ideas what's going on? It is listed in the maintained [packages list for galactic](http://repo.ros2.org/status_page/ros_galactic_default.html).
> 
> ```
> ERROR: the following packages/stacks could not have their rosdep keys resolved
> to system dependencies:
> ateam_ui: Cannot locate rosdep definition for [rosbridge_server]
> ```

Have you tried running `rosdep update` to make sure you have the latest list of keys?


We probably need to add the rosdep initialization steps to our readme since it looks like they no longer feature them in the general ROS installation instructions.
```
sudo rosdep init
rosdep update
```

guyfleeman:
Okay I did run those two steps, but it clearly failed at this point. I'll try to figure out what went wrong

ateam_ui/install_deps.sh
guyfleeman: this is definitely because I tried to skip ahead to see if it would help resolve dependencies

guyfleeman:
on a different clean 20.04 install, didn't accidentally source the bash or anything in zsh this time, getting the same error. My command sequence:

```
sudo apt update
sudo curl -sSL https://raw.githubusercontent.com/ros/rosdistro/master/ros.key -o /usr/share/keyrings/ros-archive-keyring.gpg
echo "deb [arch=$(dpkg --print-architecture) signed-by=/usr/share/keyrings/ros-archive-keyring.gpg] http://packages.ros.org/ros2/ubuntu $(. /etc/os-release && echo $UBUNTU_CODENAME) main" | sudo tee /etc/apt/sources.list.d/ros2.list > /dev/null

sudo apt update
sudo apt upgrade
sudo apt install ros-galactic-desktop ros-galactic-ros-base ros-dev-tools

source /opt/ros/galactic/setup.zsh

sudo rosdep init
rosdep update

rosdep install --from-paths . --ignore-src -y
```

fails on the last step with the above message.

guyfleeman:
this needs to be `rosdep update --include-eol-distros`. Galactic went EOL Dec 1

barulicm:
> this needs to be `rosdep update --include-eol-distros`. Galactic went EOL Dec 1

Ah, rip
Was going to talk about upgrading to Humble soon on the call tonight

guyfleeman:
Yeah your explicit belief that rosdep should have made it possible to find that package made me go back and inspect the output and cache and I saw galactic was skipped. I've removed some of the unneeded changes, but left the README.md and sanity check in setup script intact. We could just nuke this update, and attempt the upgrade tho.

barulicm:
Closed via #84

PR #86, opened 2023-01-18
Update UI Style

chachmu:
Updates the UI style to use the new ateam icon and to match the color of the top bar.

Also fixes an issue with robot placement introduced in the last ui update where the robots weren't placed centered on the center line of the field.


JNeiger (approved):

chachmu:
Ok I added a fix to this pr that makes it so you can give Line type overlays and underlays a list of points and it will make a path from them. We should be able to use this to draw trajectories so we can see what the AI is doing

PR #87, opened 2023-01-26
Setup Goalie and Shooter for Video

JNeiger:
Adds
- Adds overlay to show trajectories
- Adds the behavior planning shot + goalie behavior requests
- Adds kick ability based on distance to all robots
- Adds simple 2 step kick behavior (will be replaced)
- Adds basic PID util
- Adds basic viz singleton for trajectories
- Adds geometry angle functions
- Adds generic overlay c++ interface
- Adds kick speed + kick bool

Bugs Fixed
- Fixes quat -> yaw bug resulting in only yaws between 0 and PI
- Fixes bug with trajectory sample use (used the frame after what is expected
- Fixed angle planning bug in the trapezoidal motion profile
- Fixed acceleration sign when planning
- Fixed 0 sample trajectories if already at goal
- Unit tests for failure modes in trapezoidal motion profile
- Fixed bug where kf wasn't account for angles

Refactoring
- Cleans up the motion controller to support angle correctly and PID
- Cleans up launch files

PR #103, opened 2023-02-24
Add P controller

JNeiger:
Add P controller with support for angles

Part of #87


ateam_ai/CMakeLists.txt
```diff
@@ -42,7 +42,7 @@ add_library(${PROJECT_NAME} SHARED
   src/types/world.hpp
   src/util/message_conversions.cpp
   src/util/message_conversions.hpp
-  src/util/directed_graph.hpp
```
JNeiger, line 45: This removal is a bad git commit for this PR. It's not present in #87

PR #114, opened 2023-03-12
Joystick updates from quals

barulicm:
This PR includes changes made to the joystick controller while developing the radio bridge in preparation for qualification.

These changes include:
- Additional parameters on the joystick control node for controlling the dribbler and kicker
- A new bringup file that starts the joystick and radio bridge (coming soon) for simple manual control
- New fields in the robot motion command message for dribbler speed and kicking (will require some minor conflict resolution with other branches that added similar fields).


JNeiger (approved):
Feel free to fix the kick command as needed. The current version on `main` is very much hacked together for the video

PR #119, opened 2023-04-03
unified launch file

CollinAvidano:
I just want a unified launch file need to still update this to work on events for better and need to fix the game controller bug with symlinks being mounted into the container.


JNeiger (approved):
LGTM, I wasn't able to fully test due to wsl problems, but it seemed to launch the correct nodes (including the duplicated ssl_sim_radio_bridge_node) and the correct docker containers.

JNeiger:
@barulicm FYI: the joystick test seems to be failing in CI

barulicm:
> @barulicm FYI: the joystick test seems to be failing in CI

Hmmm... Just ran a fresh run on master and it didn't fail. Though I'm not sure how/why these changes would effect that...

barulicm:
As for this PR,

The original intent was for full_stack to basically be what we run at competition.
So, for sim testing, you might run full_stack + some simulation launch file, and for on-field testing, you might run full_stack + some field launch file.

Seeing the need people are expressing for a single launch file that starts an entire environment, perhaps we could establish a convention for top level entry point launch files? Perhaps:
- bringup_competition
  - entry point for competition
  - includes AI, UI, radio bridge, vision bridge, and gc bridge (+ any other infrastructure stuff)
- bringup_simulation
  - entry point for simulation
  - includes simulator, AI, UI, sim radio bridge, vision bridge, and gc bridge
  - Maybe includes launching the gc itself too?

That said, I would still like to maintain some organization level below these top level files, because it can be nice to keep the simulator running in between restarting the rest of the system.

barulicm:
> need to fix the game controller bug with symlinks being mounted into the container.

Wasn't aware there was a bug here. Feel free to open a bug and throw the details my way since I set up the GC integration.

CollinAvidano:
Made competiton and sim file as recommended and added args for sim bridge port ip. Merge when you can

CollinAvidano:
Adding some other stuff like restarting all nodes to this as well after talking to joe about it

hmmurdock:
reopening to get lint/CI fixed up

PR #121, opened 2023-04-04
[iLQR Refactor 1] Refactor the computation into separate files/functions

JNeiger:
Splits up our iLQR implementation into
- Problem definition (dynamics + cost)
- Calculator which optimizes the problem and gives a trajectory output
- Params struct

Not super happy with how long this PR is, but it's hard to do a partial refactor :/ Open to ideas if it's too long to review


ateam_ai/src/trajectory_generation/ilqr_computer.hpp
```diff
@@ -0,0 +1,346 @@
...
+  using CostHessian_ux = Eigen::Matrix<double, U, X>;
+  using CostHessian_uu = Eigen::Matrix<double, U, U>;
+
+  using Time = int;
+
+  using Trajectory = std::array<State, T>;
+  using Actions = std::array<Input, T>;
+
+  using Feedforwards = std::array<Input, T>;
+  using Feedbacks = std::array<Eigen::Matrix<double, U, X>, T>;
```
JNeiger, line 75: C++ Question: Is there a way to share these aliases across classes? I tried having a common templated base class that both classes inherit from, but the compiler wasn't able to find the object types.

ateam_ai/src/trajectory_generation/ilqr_computer.hpp
```diff
@@ -0,0 +1,346 @@
...
+    return h.block(0, 0, X, X);
+  }
+  inline CostHessian_ux cost_hessian_ux(const CostHessian & h)
+  {
+    return h.block(X, 0, U, X);
+  }
+  inline CostHessian_uu cost_hessian_uu(const CostHessian & h)
+  {
+    return h.block(X, X, U, U);
+  }
```
JNeiger, line 257: The problem is setup such that the jacobians/gradiants/hessians are all block matrices. Will remove in a later PR to simplify the overall code.

ateam_ai/src/trajectory_generation/ilqr_computer.hpp
```diff
@@ -0,0 +1,346 @@
...
+
+  inline Eigen::Matrix<double, U, U> inverse(const Eigen::Matrix<double, U, U> & m)
+  {
+    // Solve inverse with regulization to keep it from exploding
+    Eigen::SelfAdjointEigenSolver<Eigen::Matrix<double, U, U>> eigensolver(m);
+    if (eigensolver.info() != Eigen::Success) {
+      std::cout << "Failed to solve eigen vectors" << std::endl;
+      // TODO(jneiger): Replace with tikhonov regularization + inverse which can't fail
+      return m;
```
JNeiger, line 266: I already have the a different regularization + inverse method implemented that I didn't want to clutter this PR with. Will swap to that in a later PR so just doing the bare minimum to compile rn

JNeiger:

PR #122, opened 2023-04-04
[iLQR Refactor 2] Refactor out the problem def into the component block matricies

JNeiger:
- Remove the double computation of h_ux (h_xu)
- In the process, removes all the block operations simplifying the computer


JNeiger:

PR #123, opened 2023-04-04
[iLQR Refactor 3] Swap regularization strategies

JNeiger:
Following Kyle's suggestion about strategies.

> At first glance it seems mostly correct but I think the regularization scheme is probably the first thing to try. You're doing Quu regularization which works, but IIRC Yuval found that regularizing Vxx (by adding a small multiple of the identity matrix) is usually a bit better.


JNeiger:

PR #125, opened 2023-04-05
Adds indexed topic helpers

barulicm:
Another utility made while working on the radio bridge.

This adds a new header to ateam_common which provides `create_indexed_publishers` and `create_indexed_subscribers`. These functions create publishers / subscribers for each robot ID from a topic base name.

This PR also changes the simulation radio bridge and vision filter to use this new utility.


fourpenny:
Planning to review this later this week

fourpenny:
Decided to review this to procrastinate working on my own code...
Looks good. Maybe we can refactor the ATeamAINode to use this also since it uses the same logic you replaced in the sim radio bridge.

fourpenny (approved):

PR #126, opened 2023-04-05
Silences / addresses deprecation warnings from league protobufs

barulicm:
Disables deprecation warnings when compiling league protobuf library and removes use of deprecated fields in game controller bridge.


chachmu (approved):
LGTM 👍

PR #127, opened 2023-04-11
[iLQR] Improve convergence

JNeiger:
- Revert to the regularization in the original paper
- Use delta_v (split into 2 parts) to compute the estimated change in cost for a given alpha in the line search
- Use absolute change in cost for the convergence threshold instead of relative change in cost (May need to change this as a small alpha will result in a very small change in cost)
- Add unit tests for non-linear boundary costs for limiting acceleration magnitude


JNeiger:

PR #130, opened 2023-04-22
Use info from TeamColorListener to set team color in ateam_ai_node

fourpenny:
Part of the "Pipe ref information into the AI" task on the Comp Burndown Sheet
Takes info on the currently assigned team colors from the TeamColorListener in ateam_common and sets the robot_state_array for each callback to either our_robots or their_robots


JNeiger (approved):
LGTM!

PR #131, opened 2023-04-25
Adds hungarian algorithm

JNeiger:
Adds a matrix based interpretation of the hungarian algorithm based on the [wikipedia approach](https://en.wikipedia.org/wiki/Hungarian_algorithm#Matrix_interpretation), which is the O(N^3) implementation.

Tested through significant unit testing for each step, and then high level e2e testing using internet based examples.

Happy to split this up step by step to make the PRs smaller, but lmk

Will merge by May 9th if there are no reviews

### DIFF
--- a/ateam_common/src/assignment.cpp
+++ b/ateam_common/src/assignment.cpp
@@ -1,6 +1,9 @@
 #include "ateam_common/assignment.hpp"
 
 #include <unordered_set>
+#include <vector>
+
+#include <iostream>
 
 namespace ateam_common::assignment
 {
@@ -92,7 +95,7 @@ Eigen::MatrixXi ApplyStep3(const Eigen::MatrixXd & matrix)
 
       bool is_zero = matrix(i, j) == 0;
       if (is_zero) {
-        marks(i, j) = 1;
+        marks(i, j) = ZerosType::STARRED;
         row_covers(i) = 1;
         col_covers(j) = 1;
         break;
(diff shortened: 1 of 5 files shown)


ateam_common/src/assignment.cpp
```diff
@@ -0,0 +1,288 @@
...
+}
+
+std::optional<Eigen::Vector2i> next_uncovered_zero(
+  const CostMarkCovers & cmc)
+{
+  for (int i = 0; i < cmc.cost_matrix.rows(); i++) {
+    for (int j = 0; j < cmc.cost_matrix.cols(); j++) {
+      bool is_covered_col = cmc.col_covers(j) == 1;
+      bool is_covered_row = cmc.row_covers(i) == 1;
+      bool is_zero = cmc.cost_matrix(i, j) == 0;
```
fourpenny, line 151: You could continue to the next loop iteration after each condition here if they fail rather than checking all 3 at once. Not sure if that's more optimal in terms of speed or not.

JNeiger: True true, pulled the `is_covered_row` continue into the outer loop, and separated the `is_covered_col` continue into a difference if statement

ateam_common/src/assignment.cpp
```diff
@@ -0,0 +1,288 @@
...
+
+  do {
+    internal::apply_step_4(cmc);
+    internal::apply_step_5(cmc);
+  } while (!internal::has_unique_assignments(cmc));
+
+  std::unordered_map<std::size_t, std::size_t> assignment;
+  for (int i = 0; i < cmc.mark_matrix.rows(); i++) {
+    for (int j = 0; j < cmc.mark_matrix.cols(); j++) {
+      bool is_inbounds = i < cost_matrix.rows() && j < cost_matrix.cols();
```
fourpenny, line 48: If we go out of bounds here, we can continue to the next row since the following columns will be out of bounds; although, it seems rare to me that this would occur (when there would be fewer possible assignments than total robots), so it may not be worth it to optimize

fourpenny (approved):
Lgtm :+1: 
Implementation is sound and the math in the test cases checks out.

ateam_common/src/assignment.cpp
JNeiger: Added quick continues on the row loop and the col loop instead of just the if statement

PR #134, opened 2023-05-07
Update world in ateam_ai_node to include current game stage and command

fourpenny:
Creates a new `game_state_listener` node in `ateam_common` that can take in referee messages and return information about the current game state, the "stage" (half, penalty time, etc) and "command" (play). An instance is included as a private member of the `ateam_ai_node` that updates the `world_.referee_info` with this information every time the `timer_callback()` is called, so it can be used by the `BehaviorEvaluator`.


ateam_common/include/ateam_common/game_state_listener.hpp
```diff
@@ -0,0 +1,115 @@
...
+};
+
+class GameStateListener
+{
+public:
+  /**
+   * @brief Construct a new Game State Listener object
+   *
+   * @param node ROS node
+   * @param callback Optional callback called on changed game state - currently takes no arguments
```
barulicm, line 91: Doc copy pasta. `callback` parameter doesn't exist here.

```suggestion
```

ateam_common/src/game_state_listener.cpp
```diff
@@ -0,0 +1,89 @@
...
+  qos.transient_local();
+  ref_subscription_ = node.create_subscription<ssl_league_msgs::msg::Referee>(
+    "/gc_multicast_bridge_node/referee_messages", qos,
+    std::bind(&GameStateListener::RefereeMessageCallback, this, std::placeholders::_1));
+}
+
+void GameStateListener::RefereeMessageCallback(
+  const ssl_league_msgs::msg::Referee::ConstSharedPtr msg)
+{
+  const auto prev_command = game_command_;
```
barulicm, line 41: This variable doesn't seem to be used anywhere

```suggestion
```

ateam_common/src/game_state_listener.cpp
```diff
@@ -0,0 +1,89 @@
...
+      game_command_ = GameCommand::BallPlacementBlue;
+    // Default to Halt command
+    default:
+      game_command_ = GameCommand::Halt;
+  }
+
+  // Only check the game stage if it has changed
+  if (static_cast<GameStage>(msg->stage) != game_stage_) {
+    game_stage_ = static_cast<GameStage>(msg->stage);
+  }
```
barulicm, line 86: No need for this conditional block since you're not triggering any callbacks.

```suggestion
  game_stage_ = static_cast<GameStage>(msg->stage);
```

ateam_common/src/game_state_listener.cpp
```diff
@@ -0,0 +1,89 @@
...
+    case 13:
+      game_command_ = GameCommand::TimeoutBlue;
+    case 16:
+      game_command_ = GameCommand::BallPlacementYellow;
+    case 17:
+      game_command_ = GameCommand::BallPlacementBlue;
+    // Default to Halt command
+    default:
+      game_command_ = GameCommand::Halt;
+  }
```
barulicm, line 81: I'm curious why use a switch statement here instead of casting the value like you've done with the stage?

ateam_ai/src/ateam_ai_node.cpp
```diff
@@ -194,6 +196,11 @@ class ATeamAINode : public rclcpp::Node
       }
     }
 
+    // Get current game state for world
+    // 1. Get command
+    world_.referee_info.running_command = game_state_listener_.GetGameCommand();
+    // 2. Get the game stage
+    world_.referee_info.current_game_stage = game_state_listener_.GetGameStage();
```
barulicm, line 203: Style nit: These commands don't add any information for the reader. The function name already tells me it's getting the game command, for example.

ateam_common/include/ateam_common/game_state_listener.hpp
fourpenny: Sorry, changed this previously and forgot to push. Should be good now.

ateam_common/src/game_state_listener.cpp
fourpenny: Removed, also removed `prev_game_state_ ` since the game state is updating every time this is called now rather than on change/callback only.

ateam_common/src/game_state_listener.cpp
fourpenny: There were some depreciated commands here that I thought might cause problems with a static cast, so I added them to the enum and  am now casting the value as you suggest.

ateam_ai/src/ateam_ai_node.cpp
fourpenny: Kept the brief "Get current game state for world", but removed the subsections, since the other code in this node has similar comments.

ateam_common/include/ateam_common/game_state_listener.hpp
barulicm: hmmm nothing seems to have changed here...

barulicm (approved):

JNeiger (approved):

PR #136, opened 2023-05-08
Adds radio bridge

barulicm:
It's finally here!

This PR adds the `ateam_radio_bridge` package. This package provides a node to connect our robot radio network to ROS.

This package also includes a submodule for the [software-communication repo](https://github.com/SSL-A-Team/software-communication), which holds the shared definitions of our radio packets.


hmmurdock (approved):
This code looks to be higher quality than any C++ I could write and it works on the robot. Approved.

PR #137, opened 2023-05-12
Significant rewrite of the UI

### LINKED ISSUE #91: Improve UI performance
The UI uses 100% of CPU and uses and increasingly large amount of memory, likely due to rerending the DOM on every robot state update.

From slack:
I been trying to debug the UI performance and I think I've narrowed it down some. The constant updates to the robot array is causing the re-rendering of the robot's to fully use all the CPU as well as causing a memory "leak". This results in an every increasing delay between getting the info from ROS2 and displaying it.

First image is of the Vue debugger after like 10-15 minutes. Second image is of the chrome performance tab. This is just idle display. Note the essentially maxed out CPU usage + the ever increasing heap usage. I've had the tab crash due to OOM before.

I removed the robot rendering and at most, it's 100 ms of scripting over 20 seconds from the on-mouse-movement callback. At 100hz + 16 * 2 robots, that's 3200 updates a second (assuming they actually batch the updates correctly). Once I started trying to draw the trajectories as well at 100hz, it all went downhill lol.

chachmu:
Revamps the existing UI to use typescript and replaces the Konva canvas management library with PixiJS which seems to have much better performance and support. 

Also adds a UI element to set goalie id, robot status indicators for the breakbeam detection and motor errors, and field dimension updates from vision.

Closes #91


ateam_ui/src/src/robot.ts
```diff
@@ -0,0 +1,166 @@
+import { TeamColor } from "@/team"
+import { RenderConfig } from "@/state"
+import * as PIXI from "pixi.js"
+import ROSLIB from "roslib"
+
+// Types used for team data, our team and the opponent team intentionally both use the same class
+// with a place for data about robot status, active plays, etc so that we can use a single UI to control
+// two instances of our software playing against each other for testing
```
barulicm, line 8: Copied from team.ts?

barulicm (approved):

PR #139, opened 2023-05-29
Add params for bridges

CollinAvidano:
Add parameters for the ports and ips of the two bridges as was in the software burndown.


ateam_bringup/launch/ssl_grsim.launch.xml
```diff
@@ -1,8 +1,13 @@
 <launch>
   <arg name="headless" default="true"/>
+ <arg name="ssl_sim_radio_ip" default="224.5.23.2"/>
```
barulicm, line 3: Was this meant to be 127.0.0.1 as is the default in the code?
```suggestion
 <arg name="ssl_sim_radio_ip" default="127.0.0.1"/>

```

barulicm (approved):

PR #140, opened 2023-06-09
Add a more advanced assignment

JNeiger:
Adds an assignment algorithm that takes into account priority

Assumptions to speed up implementation

- The depth of the DAG(t) is not too deep such that it "eats" all the robots available
- The optimization cost of time is equal to the time of the last frame of the trajectory planned
- All required plans have required parents OR are root nodes
- Non-required nodes have no children

Features

- Priority based globally optimized assignment
- Time based cost function. Directly measures the quantity we care about and can easily be balanced if we add other cost (like robot assignment change costs, we can say change robots if we can get there more than 2 seconds faster than now)
- Fixed ATEAM_CHECK with multiple includes
- Fixes non-square->square, rows must be <= cols

Todo:

- [x] Final unit tests at the highest levels
- [x] More unit tests for different edge cases (non-square assignment, empty input)
- [x] Test with rest of stack
- [x] Confirm ATEAM_CHECK's actually print out a good stack trace

Most of the implementation will not change at the "lower" levels, but putting this up for review now so the iteration time is lower


fourpenny:
Not sure why this is happening but I'm seeing an inconsistency with the number of tests running when I call `colcon test` vs. the build and test CI pipeline here.
image

JNeiger:
What version of ros2 are you running? I think we updated earlier this year and that was a cause of different "broken" style tests. There's a --show_results option (or something) that will at least say if it's a style or unit test issue

ateam_common/include/ateam_common/status.hpp
```diff
@@ -17,7 +17,7 @@
 // LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 // THE SOFTWARE.
-
+#pragma once
```
barulicm, line 20: Redundant pragma once when we already have the header guards below?

ateam_common/include/ateam_common/status.hpp
```diff
@@ -137,6 +127,7 @@ StatusOr<T> Failure(std::string && s)
       if (!__status_or_lhs) { \
         std::stringstream s; \
         s << text << "\n" << __status_or_lhs.error() << std::endl; \
+        std::cout << s.str(); \
```
barulicm, line 130: Is this a leftover debug printout?
If not, should it print to `std::cerr` instead?

fourpenny:
Okay, here is what I understand about what we have here based on your code and the slides on our software design:
0. The BehaviorEvaluator chooses the best DAG(t) of BehaviorGoals to assign and execute based on whatever criteria we decide on later
1. The BehaviorRealization creates a series of trajectories to realize these goals based on the current world/robot states
2. We then map each goal node in the DAG(t) to a BehaviorPlan
- Right now, we're just generating plans with empty trajectories except for their `.time` value.
4. Each of these BehaviorPlans are assigned to a robot using assignment
- We do this by priority group: For each priority group, which we have the node indices for, we assign trajectories to robots using `assignment`, doing this one node at a time for the Required group and for multiple nodes in the other groups.
5. We return a DAG(t) of BehaviorPlans that we can then execute within the BehaviorExecuter

Just want to verify, is this accurate?
Seems like the assignment here for non-square matrices is okay as long as we follow the assumptions you state here relating to parent/child nodes for required/non-required behaviors when constructing our graphs of BehaviorGoals.

fourpenny:
This is what I get as output when I run the test commands you suggested:
image
[test_output.txt](https://github.com/SSL-A-Team/software/files/11780014/test_output.txt)

fourpenny:
The failed tests I'm running into aren't related to the behavior stuff- looks like they are related either to Python linting (flake8) or the joystick control node for some reason.

JNeiger:
Yep! That's exactly the flow. The only difference is that we do generate "real" plans for one of the kicks and the move behavior goal. Both do not avoid collisions and just use a simple trapezoidal velocity profile to reach the goals. One thing to note is that the "kick" action doesn't plan the full trajectory from current location until the actual kick, so the time estimate for that will be an underestimate of the true goal completion time. This may result in some weird assignments near the ball.

JNeiger:
Hmmmm, I'm really not sure what's up with the joystick or launch file stuff. I thought the launch file changes were reverted? And the joystick stuff hasn't been updated in a while :/

@barulicm may have some ideas, but I'm fine just using the CI as the source of truth

ateam_common/include/ateam_common/status.hpp
JNeiger: Will change to std::cerr

ateam_common/include/ateam_common/status.hpp
JNeiger: Will remove

barulicm:
wrt testing discrepancies, it may be related to the fact that CI runs the tests on the merge result of the PR branch with main. If you just checkout this branch and run the tests, that's not the same code the CI is running against. This PR's branch is currently 57 commits behind main.

fourpenny (approved):
I think I'm okay with this merging since the automated CI/CD seems to be working and my personal issues seem to be with other packages anyway. I'll continue to work on getting my personal copy of the repo to build and test properly.

fourpenny:
Yes, sorry, we do create real trajectories for a few things in the codebase- the flow I mentioned above specifically was based off of reading the changed source files and the behavior executor test.

> Yep! That's exactly the flow. The only difference is that we do [generate "real" plans](https://github.com/SSL-A-Team/software/blob/main/ateam_ai/src/trajectory_generation/trajectory_generation.cpp) for one of the kicks and the move behavior goal. Both do not avoid collisions and just use a simple trapezoidal velocity profile to reach the goals. One thing to note is that the "kick" action doesn't plan the full trajectory from current location until the actual kick, so the time estimate for that will be an underestimate of the true goal completion time. This may result in some weird assignments near the ball.

barulicm (approved):

PR #141, opened 2023-06-13
field geometry and coordinate system handling redux

CollinAvidano:
Implementing new plan Barulic and I had talked about to just have all field geometry be published from where we already subscribe to it in the vision filter. Adds inverting logic to all vision estimates before they go into the vision filter depending on what team we are. To do that also adds a side listener into the color info listener since its dependent and we need it most of the same places (Joe agreed).


ateam_vision_filter/src/types/robot_measurement.hpp
```diff
@@ -27,6 +27,34 @@ struct RobotMeasurement
...
+    return angle_rad;
+  }
+
+  double angle_wrap_pi(double angle_rad)
+  {
+    if (angle_rad < -M_PI || M_PI < angle_rad) {
+      return angle_wrap_2pi(angle_rad + M_PI) - M_PI;
+    }
+    return angle_rad;
+  }
```
barulicm, line 57: These functions are already available in ROS's angles package.
See [`angles::normalize_angle_positive()`](http://docs.ros.org/en/humble/p/angles/generated/function_angles_8h_1a100ca493044236d7e099e08e78aab5bf.html) and [`angles::normalize_angle()`](http://docs.ros.org/en/humble/p/angles/generated/function_angles_8h_1a90ae8baf394c3ce775a3a14e48c7e7be.html).

ateam_vision_filter/src/types/robot_measurement.hpp
```diff
@@ -27,6 +27,34 @@ struct RobotMeasurement
 {
   Eigen::Vector2d position;
   double theta;
+
+  void invert()
+  {
+    position *= -1.0;
+    theta = angle_wrap_pi(theta + M_PI);
```
barulicm, line 34: It looks like these inversion functions are rotating the field 180deg (as opposed to just flipping X).
Is this intentional? (not saying it's wrong, just checking if we decided to change this since the in person conversation about this)

CollinAvidano: Yeah that was intentional I was talking to Joe and Woodward about the plan and I think the worry was if you just flip x left and right rotations are inverted. But there may be other benefits im thinking of to just flipping x that im not thinking about.

ateam_vision_filter/src/types/robot_measurement.hpp
CollinAvidano: Sorry im just really used to my stupid new companies minimal depedencies rule. Its been rough....

barulicm (approved):

PR #142, opened 2023-06-17
Fixes member name to match new GC API

barulicm:
The GC got a big update recently. One of the member names in the protobuff API changed. This broke our UI's GC shortcut buttons.
This PR fixes the member name in our code to match the new name in the GC.


fourpenny:
lgtm

fourpenny (approved):
still lgtm

PR #143, opened 2023-06-19
Basic geometry package

fourpenny:
Adds a new `ateam_geometry` package with basic utilities for creating and doing operations with the following:
- Line segments
- Circles
- Rectangles

All of the "points" used for these utilities are `Eigen::Vector2d` objects.

I'm planning to implement some basic tests for these before merging to check my logic with some examples that I can calculate by hand and that represent some edge cases, so think of this as a draft as of today (6/18). I put this up so hopefully people can 1. roast my fledgling C++ skills 2. catch any easy to see logical errors and 3. suggest important stuff I've missed in terms of functionality.

I screwed up when merging this with main, so I'll be fixing that tomorrow (6/19).


ateam_geometry/src/segment.cpp
```diff
@@ -0,0 +1,169 @@
...
+
+        Eigen::Vector2d p_relative; // projection of P onto A
+        p_relative.x() = point.x() - segment.p1.x();
+        p_relative.y() = point.y() - segment.p1.y();
+        double projection = direction.dot(p_relative);
+        
+        if (projection < 0 || projection > segment.length) {
+            return false;
+        }
+        return true;
```
JNeiger: We also need to check that the distance between the projection point and the input point is 0 I think

JNeiger (commented):
Sorry about the dump of comments, happy to jump in a call or something if needed

ateam_geometry/src/rectangle.cpp
```diff
@@ -0,0 +1,78 @@
...
+
+namespace ateam_geometry {
+    Rectangle::Rectangle(const Eigen::Vector2d & v1, const Eigen::Vector2d & v2) {
+        width = std::abs(v1.x() - v1.y());
+        height = std::abs(v1.y() - v2.y());
+        
+        corners[0] = v1;
+        corners[1] = v2; 
+        corners[2] = Eigen::Vector2d(v1.x(), v2.y());
+        corners[3] = Eigen::Vector2d(v2.x(), v1.y());
```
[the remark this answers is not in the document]
fourpenny: For the MVP, I'm going to keep this simple since I know it would be a few more calculations to ensure this. Let's add this as a TODO for v2?

ateam_geometry/include/ateam_geometry/utilities.hpp
```diff
@@ -0,0 +1,26 @@
...
+// THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+#include <Eigen/Dense>
+#include <math.h>
+
+namespace ateam_geometry {
+    double cross_product_2d(const Eigen::Vector2d & w, const Eigen::Vector2d & v);
```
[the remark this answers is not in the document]
fourpenny: I *think* this is only implemented in Eigen for 3D and complex numbers, which is a slightly different implementation than the one provided here.

ateam_geometry/src/segment.cpp
fourpenny: Isn't it just on the line (on the first endpoint) in that case, which would return true?

ateam_geometry/src/rectangle.cpp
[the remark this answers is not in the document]
fourpenny: Are there any quick sorting functions for this in C++? I did this in a Python prototype using:
```
self.corners = sorted([
            Point(v1.x, v1.y),
            Point(v1.x, v2.y),
            Point(v2.x, v2.y),
            Point(v2.x, v1.y)
        ], key= attrgetter('x', 'y'))
```

ateam_geometry/src/segment.cpp
```diff
@@ -0,0 +1,169 @@
...
+#include "ateam_geometry/utilities.hpp"
+#include <Eigen/Dense>
+#include <math.h>
+#include <optional>
+#include <tuple>
+
+namespace ateam_geometry {
+    LineSegment::LineSegment(const Eigen::Vector2d & start, const Eigen::Vector2d & end) {
+        p1 = start;
+        p2 = end;
```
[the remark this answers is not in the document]
fourpenny: I changed the naming conventions to make this non-directional, but I can change this to do the opposite if you want.

ateam_geometry/src/rectangle.cpp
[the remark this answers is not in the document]
JNeiger: I'm not sure there's a good way with sorting that would order it in the correct way. My first thought is something like the following to get counter clockwise starting at the bottom left
```
corners.at(0) = Eigen::Vector2d{std::min(v1.x(), v2.x()), std::min(v1.y(), v2.y())};
corners.at(1) = Eigen::Vector2d{std::min(v1.x(), v2.x()), std::max(v1.y(), v2.y())};
corners.at(2) = Eigen::Vector2d{std::max(v1.x(), v2.x()), std::max(v1.y(), v2.y())};
corners.at(3) = Eigen::Vector2d{std::max(v1.x(), v2.x()), std::min(v1.y(), v2.y())};
```

ateam_geometry/src/segment.cpp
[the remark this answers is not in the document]
JNeiger: I'm happy either way, we should just be consistent

ateam_geometry/include/ateam_geometry/utilities.hpp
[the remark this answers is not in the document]
JNeiger: Oh, you're totally right. I completely missed the [footnote ](https://eigen.tuxfamily.org/dox/group__Geometry__Module.html#ga0024b44eca99cb7135887c2aaf319d28)`Remember that cross product is only for vectors of size 3` :/


While googling around for this, apparently it's only defined for a 3 and [7 dimensional](https://en.wikipedia.org/wiki/Seven-dimensional_cross_product) vector?? wtf math lol

ateam_geometry/src/segment.cpp
JNeiger: image

I don't think that's guaranteed. Only blue in the image above should return true, but I think the current code will allow blue or green to return true. Specifically for green, we will find the projection of green onto the segment and just see if that projection is within the bounds of the segment length. Green can be thousands of meters away, but as long as it's perpendicular to the segment, it will return true I think.

See the last 6 paragraphs of the answer [here](https://computergraphics.stackexchange.com/a/2108)

ateam_geometry/include/ateam_geometry/utilities.hpp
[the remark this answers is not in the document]
fourpenny: I'm so curious as to how someone figured out the properties of the 7 dimensional cross product, and maybe more importantly, why they needed to know that

ateam_geometry/src/segment.cpp
fourpenny: Sorry, I think I misunderstood your original comment and was thinking you were referring to changing line 88. This check makes sense

fourpenny: Okay, I think this should be better now.

ateam_ai/src/behavior/behavior_evaluator.cpp
```diff
@@ -1,3 +1,22 @@
+// Copyright 2023 A Team
```
chachmu, line 1: Did we mean to have our copyright twice in a row in this file?

ateam_geometry/src/segment.cpp
fourpenny: Jk that was very wrong but maybe it's better now?

ateam_ai/src/behavior/behavior_evaluator.cpp
fourpenny: Nope, I'll remove it

ateam_geometry/src/segment.cpp
fourpenny: This should be tested and work

ateam_geometry/src/rectangle.cpp
```diff
@@ -0,0 +1,94 @@
...
+{
+  width = std::abs(v1.x() - v2.x());
+  height = std::abs(v1.y() - v2.y());
+
+  // TODO(Christian): Ensure standard order of corners, with
+  // min x/y at index 0.
+  corners.at(0) = v1;
+  corners.at(1) = v2;
+  corners.at(2) = Eigen::Vector2d(v1.x(), v2.y());
+  corners.at(3) = Eigen::Vector2d(v2.x(), v1.y());
```
CollinAvidano, line 36: Yeah we will need to standardize this because because

ateam_geometry/src/circle.cpp
```diff
@@ -0,0 +1,56 @@
...
+    Eigen::Vector2d point =
+      LineSegment(center, radius, offset + (i * spacing)).p2;
+    points.push_back(point);
+  }
+  return points;
+}
+
+bool is_point_in_circle(const Eigen::Vector2d & point, Circle & circle)
+{
+  return (point - circle.center).norm() <= circle.radius;
```
CollinAvidano, line 54: We may want more than this such as intersection points with circle is another common operation we need but if you dont want to add it now or dont think we will need it we dont have to.

ateam_geometry/src/segment.cpp
```diff
@@ -0,0 +1,232 @@
...
+  return points;
+}
+
+bool is_point_on_segment(const Eigen::Vector2d & point, LineSegment & segment, double tolerance)
+{
+  /*
+      https://computergraphics.stackexchange.com/questions/2105/test-if-a-point-is-on-a-line-segment
+      https://lucidar.me/en/mathematics/check-if-a-point-belongs-on-a-line-segment/
+  */
+  double crossProduct = (point.y() - segment.p1.y()) * (segment.p2.x() - segment.p1.x()) -
```
CollinAvidano, line 90: I mean you dont have to but you have a cross product func you could call twice for this

ateam_geometry/src/segment.cpp
```diff
@@ -0,0 +1,232 @@
...
+    (point.x() - segment.p1.x()) * (segment.p2.y() - segment.p1.y());
+
+  if (std::abs(crossProduct) > tolerance) {
```
CollinAvidano, line 93: Unsure this tolerance should be used for both the dot prod and the cross product.

CollinAvidano (approved):
I left a few non essential comments for documentation for later but I think this is good to merge for now

PR #148, opened 2023-06-23
Add B Splines

JNeiger:
Still need to finish up the control point -> trajectory API and the sampling algorithm of the curve itself.

### DIFF
--- a/ateam_ai/src/trajectory_generation/b_spline.cpp
+++ b/ateam_ai/src/trajectory_generation/b_spline.cpp
@@ -1,0 +1,117 @@
+// Copyright 2021 A Team
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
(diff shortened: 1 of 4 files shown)


ateam_ai/src/trajectory_generation/b_spline.cpp
```diff
@@ -0,0 +1,117 @@
...
+    out.push_back(control_points.front());
+  }
+  for (int i = 1; i < control_points.size() - 1; i++) {
+    out.push_back(control_points.at(i));
+  }
+  for (int i = 0; i < multiplicity; i++) {
+    out.push_back(control_points.back());
+  }
+
+  return out;
```
barulicm, line 36: Consider an algorithm-based implementation
```suggestion
  std::vector<Eigen::Vector2d> out(control_points);
  std::fill_n(std::front_inserter(out), multiplicity, control_points.front());
  std::fill_n(std::back_inserter(out), multiplicity, control_points.front());
  return out;
```

ateam_ai/src/trajectory_generation/b_spline.cpp
```diff
@@ -0,0 +1,117 @@
...
+  return constant_spacing(control_points);
+}
+
+std::vector<double> constant_spacing(const std::vector<Eigen::Vector2d> & control_points) {
+  std::vector<double> out;
+  double spacing = 1.0 / (control_points.size() - 1);
+  for (int i = 0; i < control_points.size(); i++) {
+    out.push_back(i * spacing);
+  }
+
```
barulicm, line 49: Consider an algorithm-based implementation
```suggestion
  std::ranges::transform(control_points,
                         std::back_inserter(out),
                         [&spacing](const auto& in_point){
                             return in_point * spacing;
                         });
```

barulicm (approved):

ateam_ai/src/trajectory_generation/b_spline.cpp
JNeiger: Rip, std::front_inserter requires push_front that vector doesn't have. I'll add a todo for later and leave it as is for now. The code is no where near performant as is

ateam_ai/src/trajectory_generation/b_spline.cpp
JNeiger: Hitting some template compile errors that I don't think I have enough to really solve. Gonna leave as is with a todo as well

PR #151, opened 2023-06-27
RRT Path Planner

barulicm:
Adds an [OMPL](https://ompl.kavrakilab.org/)-based RRT path planner for future use in AI code.


barulicm:
@JNeiger I'm opening this draft to make sure we're on the same page on the interface / general approach. Still want to add some more tests and inspect the output trajectories more closely to make sure they're reasonable. (I haven't actually visualized any of the paths this generates yet.)

JNeiger:
Interface lgtm. I didn't look too deep into ompl's interface, but it looks like it has a start / end point at minimum (if the path is valid) which is all I need.

ateam_bringup/launch/bringup_simulation.launch.py
```diff
@@ -59,7 +59,7 @@ def generate_launch_description():
 
     ui_path = os.path.join(get_package_share_directory('ateam_ui'), 'launch')
     ui_launch = launch.actions.IncludeLaunchDescription(
-        PythonLaunchDescriptionSource([ui_path, '/ateam_ui_launch.py']))
+        PythonLaunchDescriptionSource([ui_path, '/ateam_ui_debug_launch.py']))
```
barulicm, line 62: This is an accidental change, but everyone in the hotel room seemed to like it, so I'll keep it in this PR.

JNeiger (approved):

PR #152, opened 2023-06-28
Add robot constants

fourpenny:
Created a very basic header file in ateam_common with robot constants. Let me know if there is something I missed or if we want to rename this file and/or include general constants.


chachmu (approved):
:+1:

PR #154, opened 2023-06-30
Plays: defense, halt, stop

fourpenny:
Has some useful bits for generating stop compliant plays (including speed limiting), halting the robots (including a new BehaviorGoal type), and generating semi-logical defensive movements to block the ball.

I'll keeping adding to this as needed, but I'm about to get on my plane to France so I thought I'd make a PR in case anyone else wants to use it while I'm busy.


ateam_ai/src/plays/defense.hpp
```diff
@@ -0,0 +1,125 @@
...
+    ateam_geometry::Point(
+      -4,
+      0.5), ateam_geometry::Point(
+      -4, -0.5));
+
+  // Get the ball location
+  std::optional<Eigen::Vector2d> ball_location = world.get_unique_ball();
+  while (!ball_location.has_value()) {
+    ball_location = world.get_unique_ball();
+  }
```
JNeiger, line 66: Get unique ball should either be the actual ball or nothing. Nothing means there is no ball. I think this would be an infinite loops if we have no ball.

ateam_ai/src/plays/defense.hpp
```diff
@@ -0,0 +1,125 @@
...
+}
+
+BehaviorGoal get_goalie_behavior_goal(const FieldSidedInfo & our_side_info)
+{
+  // Line that is 0.5 m from the defense area in all directions
+  ateam_geometry::Segment goalie_line = ateam_geometry::Segment(
+    ateam_geometry::Point(
+      -4,
+      0.5), ateam_geometry::Point(
+      -4, -0.5));
```
JNeiger, line 60: We'll have to confirm the field size is correct for div b if we haven't already. Grsim may default to div a

ateam_ai/src/plays/defense.hpp
```diff
@@ -0,0 +1,125 @@
...
+  const World & world, const Field & field,
+  const int & num_defenders)
+{
+  std::vector<BehaviorGoal> defenders;
+  std::optional<Eigen::Vector2d> ball_location = world.get_unique_ball();
+  // Get line between the ball and the goal
+  while (!ball_location.has_value()) {
+    ball_location = world.get_unique_ball();
+  }
+  ateam_geometry::Point ball = ateam_geometry::EigenToPoint(ball_location);
```
JNeiger, line 89: Same with the whole loop here

ateam_ai/src/plays/defense.hpp
```diff
@@ -0,0 +1,125 @@
...
+        if (ateam_geometry::Segment(
+            previous_point,
+            candidate).squared_length() > pow(robot_diameter, 2))
+        {
+          BehaviorGoal go_to_point {
+            BehaviorGoal::Type::MoveToPoint,
+            BehaviorGoal::Priority::Required,
+            MoveParam(ateam_geometry::PointToEigen(candidate))
+          }
+          points_around_ball.push_back(go_to_point);
```
JNeiger, line 113: I'm on mobile so I might be missing it, but I don't think points around ball is defined

ateam_ai/src/plays/stop.hpp
```diff
@@ -0,0 +1,111 @@
...
+    std::vector<ateam_geometry::Point> candidate_points;
+    candidate_points.reserve(50);
+    std::copy_n(circlePoints, 50, std::back_inserter(candidate_points));
+    // Pick the first points that are within bounds on the circle
+    // and a robot's diameter away from each other
+    ateam_geometry::Point previous_point = ateam_geometry::Point(-100, -100);
+    for (ateam_geometry::Point candidate : candidate_points) {
+      if (is_point_in_bounds(candidate)) {
+        if (ateam_geometry::Segment(
+            previous_point,
```
JNeiger, line 92: Previous point is never updated with the previous point, it's always -100, -100 I think

ateam_ai/src/plays/defense.hpp
fourpenny: I think there are given situations where we may not have a ball yet (or not have filtered one yet) but may want to call a play anyway... i.e., if the referee is moving the ball and it can't be seen properly by vision. So maybe we can call this a limited number of times (say 10ish max?) and then if it fails too much, we return an empty vector?

ateam_ai/src/plays/defense.hpp
fourpenny: As long as our coordinates are in meters, I think this is correct:
https://robocup-ssl.github.io/ssl-rules/sslrules.html#_dimensions
(The defense area is [3.5, 4.5] m in the x direction and [-1, 1] in the y direction on each team's side on the Div B field)

ateam_ai/src/plays/defense.hpp
fourpenny: Addressed this in the same way as the other one

ateam_ai/src/plays/defense.hpp
fourpenny: I copied and pasted some code here and forgot to change variable names to match. Should be correct now (using `defenders` vector)

ateam_ai/src/plays/stop.hpp
fourpenny: Added statement to update this in the loop

ateam_ai/src/plays/defense.hpp
fourpenny: > As long as our coordinates are in meters, I think this is correct: https://robocup-ssl.github.io/ssl-rules/sslrules.html#_dimensions (The defense area is [3.5, 4.5] m in the x direction and [-1, 1] in the y direction on each team's side on the Div B field)

So for this function, we simply move 0.5 m from each of those boundaries :)

ateam_ai/src/plays/defense.hpp
JNeiger: I think that's the best we can do. Especially while the trajectory follower controller is still in the ai module. Once that is removed, we can get away with more weird cycle times

ateam_ai/src/plays/defense.hpp
JNeiger: Ezpz

ateam_ai/src/plays/defense.hpp
```diff
@@ -0,0 +1,129 @@
...
+#include <array>
+#include <vector>
+#include <optional>
+
+#include "util/directed_graph.hpp"
+#include "types/world.hpp"
+#include "types/behavior_goal.hpp"
+#include "ateam_geometry/ateam_geometry.hpp"
+
+const FieldSidedInfo & our_side_info
```
CollinAvidano, line 35: Does this build?

ateam_ai/src/plays/defense.hpp
```diff
@@ -0,0 +1,129 @@
...
+  defense_graph.add_node(goalie);
+
+  std::vector<BehaviorGoal> defenders = get_defense_behavior_goals(world, 2);
+  for (BehaviorGoal defender : defenders) {
+    defense_graph.add_node(defender);
+  }
+  return defense_graph;
+}
+
+BehaviorGoal get_goalie_behavior_goal(const World & world, const FieldSidedInfo & our_side_info)
```
CollinAvidano, line 52: const FieldSidedInfo & our_side_info is contained in world

CollinAvidano (commented):
Yeah I dont know these arent that critical I feel the goalie box one is a bit. Other than that im fine with merging.

PR #158, opened 2023-07-22
Add heat map to UI

CollinAvidano:


ateam_ui/src/src/field.ts
```diff
@@ -30,7 +30,28 @@ export class Field {
 
     constructor() {
         this.fieldDimensions = new FieldDimensions();
+        // this.overlays = [];
```
chachmu, line 33: We probably don't need this comment

ateam_ui/src/src/field.ts
```diff
@@ -30,7 +30,28 @@ export class Field {
...
+                   {mesh1d: [1,4,5,5,6,7,8,1,7,4,14,1,34,1]},
+                   {mesh1d: [1,4,5,40,6,7,8,1,7,4,14,1,34,1]},
+                   {mesh1d: [1,4,14,5,6,7,8,1,3,4,14,1,34,1]},
+                   {mesh1d: [1,4,14,5,6,7,8,1,7,4,14,1,34,1]},
+                   {mesh1d: [1,4,14,5,6,7,8,1,7,4,14,1,34,1]},
+                   {mesh1d: [1,4,5,5,6,7,8,1,7,4,14,1,34,1]}
+                  ]
+          }
+        )
+      )
```
chachmu, line 54: We also probably don't want to add the test shader to the codebase

ateam_ui/src/src/overlay.ts
```diff
@@ -123,12 +168,57 @@ export class Overlay {
                 break;
             // MESH
             case 6:
-                // I think I can use a PIXI filter to do this more efficiently
+                // This is for 2d arrays of meshes
+                // ASSUMPTION MESH IS IN Y then X or row then column (row-major)
+                // Could just use flat if this was directly an array of arrays
+                let heat_texture = this.array_to_pixi_texture(this.mesh, this.mesh_alpha)
+                // console.log(heat_texture.valid);
```
chachmu, line 175: We should probably remove all the unused/commented out lines if they aren't necessary

ateam_ui/src/src/overlay.ts
```diff
@@ -1,6 +1,7 @@
 import { RenderConfig } from "@/state"
 import { Field } from "@/field"
-import * as PIXI from "pixi.js";
+import * as PIXI from "pixi.js"
+import { Texture } from "pixi.js"
```
chachmu, line 4: Is there a reason for this import? It seems like all the code still uses PIXI.Texture so it should be covered by the line above this one

ateam_ui/src/src/overlay.ts
```diff
@@ -34,6 +36,49 @@ export class Overlay {
...
+                const float e1 = 0.125;
+                const vec4 v1 = vec4(0,0.23529411764705882,0.6666666666666666,1);
+                const float e2 = 0.375;
+                const vec4 v2 = vec4(0.0196078431372549,1,1,1);
+                const float e3 = 0.625;
+                const vec4 v3 = vec4(1,1,0,1);
+                const float e4 = 0.875;
+                const vec4 v4 = vec4(0.9803921568627451,0,0,1);
+                const float e5 = 1.0;
+                const vec4 v5 = vec4(0.5019607843137255,0,0,1);
```
chachmu, line 52: Where are all these constants coming from? Should we have them be some sort of config/variable?

ateam_ui/src/src/overlay.ts
```diff
@@ -123,12 +168,57 @@ export class Overlay {
...
+            }
+          }
+
+          return PIXI.Texture.fromBuffer(buff, width, height, {scaleMode: PIXI.SCALE_MODES.LINEAR	});
+        }
+      }
+
+      // Regardless of what we return the user can just check if this.valid is true
+      return PIXI.Texture.fromBuffer(null, 0, 0);
+    }
```
chachmu, line 222: It seems like we suddenly switch from 4 space indents to 2 space indents, we should probably keep it consistent

ateam_ui/src/src/state.ts
```diff
@@ -118,7 +118,7 @@ export class AppState {
 
     getOverlayCallback() {
 	    const state = this; // fix dumb javascript things
-	    return function(msg:any) {
+      return function(msg:any) {
```
chachmu, line 121: I think this also needs 2 more spaces in the indent

chachmu (changes requested):
We should probably also wait until the rest of the UI is merged into main to merge this PR.

ateam_ui/src/src/overlay.ts
CollinAvidano: You know how I said it was a nightmare trying to get different Cmaps and interpolation schemes from that one glsify package to import correctly? This is the hard coded default from that. Pretty sure we talked about this.

CollinAvidano: We can separate them into a file if you want and just read that file into the string later. Notice it's a variable number of stops. We would need to write what glsify does for auto generating these or just get it working

ateam_ui/src/src/overlay.ts
CollinAvidano: Nope just need to remove that

ateam_ui/src/src/field.ts
CollinAvidano: Yeah I kinda left that there so you could see format

ateam_ui/src/src/overlay.ts
CollinAvidano: I would say I will get rid of the log but I should put it as a doc somewhere what the format is or else someone is going to get vonfused

ateam_ui/src/src/overlay.ts
```diff
@@ -123,12 +168,57 @@ export class Overlay {
...
             case 6:
-                // I think I can use a PIXI filter to do this more efficiently
+                // This is for 2d arrays of meshes
+                // ASSUMPTION MESH IS IN Y then X or row then column (row-major)
+                // Could just use flat if this was directly an array of arrays
+                let heat_texture = this.array_to_pixi_texture(this.mesh, this.mesh_alpha)
+                // console.log(heat_texture.valid);
+                if (heat_texture.valid) {
+                  this.heat_filter.uniforms.uTexture = heat_texture;
+                //   this.heat_filter.uniforms.uSample = heat_texture;
```
CollinAvidano, line 178: Yeah this I need to get rid of

ateam_ui/src/src/overlay.ts
chachmu: The descriptive comments are fine, I was referring to lines of code that have been commented out

chachmu (approved):
:+1:

Issue #160, opened 2023-10-08
Merge Comp2023 changes


barulicm:
current plan is to go package-by-package in this order:

- [x] ateam_geometry
- [x] ateam_msgs
- [x] ateam_common
- [x] ateam_ui
- [x] ateam_game_controller_bridge
- [x] ateam_joystick_control
- [x] ateam_kenobi
- [x] ateam_radio_bridge
- [x] ateam_ssl_simulation_radio_bridge
- [x] ateam_ssl_vision_bridge
- [x] ateam_vision_filter
- [x] ateam_bringup

barulicm:
ateam_geometry: #164 
ateam_msgs: #161 
ateam_ui: #163

barulicm:
ateam_joystick_control: #165

barulicm:
ateam_game_controller_bridge: #171
ateam_kenobi: #170

barulicm:
ateam_radio_bridge: #177 
ateam_ssl_simulation_radio_bridge: #178 
ateam_ssl_vision_bridge: #179 
ateam_vision_filter: #180 
ateam_bringup: #181

barulicm:
All related PRs have been merged :tada:

PR #164, opened 2023-10-08
Imports comp/2023 changes for ateam_geometry

barulicm:
Just a few functions and typedefs added with basic tests.


ateam_geometry/test/normalize_test.cpp
fourpenny, line 1: Math checks out according to the GNOME calculator

fourpenny (approved):
Test math checks out and it's not broken

fourpenny:
Merged as part of https://github.com/SSL-A-Team/software/issues/160

PR #161, opened 2023-10-08
Comp 2023: ateam_msgs

chachmu:
Everything seems to be working , we mostly added fields instead of renaming/deleting anything.

PR #165, opened 2023-10-08
Comp 2023: ateam_joystick_control

barulicm:
Pulls in the ateam_joystick_control changes from competition.

Ready to review once #161 is merged.


ateam_joystick_control/src/joystick_control_node.cpp
```diff
@@ -88,19 +91,35 @@ class JoystickControlNode : public rclcpp::Node
...
-    const auto placeholder_pos = topic.find("{}");
-    if (placeholder_pos != std::string::npos) {
-      topic.replace(placeholder_pos, 2, std::to_string(robot_id));
+    if (robot_id == -1) {
+      control_publisher_.reset();
+    } else {
+      std::string topic = command_topic_template_;
+      const auto placeholder_pos = topic.find("{}");
+      if (placeholder_pos != std::string::npos) {
+        topic.replace(placeholder_pos, 2, std::to_string(robot_id));
```
fourpenny, line 110: If there's a placeholder, it's just at the beginning of the string, I assume from the ending string position (2)? Why not use the variable we assign on line 108 + 2 instead of an absolute/magic number?

fourpenny (commented):
~Should be fine, just curious about the one question above ^~ Reread docs and was clear. LGTM!

fourpenny (approved):

chachmu (approved):
:+1: LGTM

PR #166, opened 2023-10-08
Comp 2023: ateam_common package

fourpenny:
Adds in changes to the ateam_common package made for Comp 2023, with minimal adjustments to some other packages (radio_bridge, ai, and vision_filter) to make sure the code builds and tests pass.

This PR needs to go in before many of the others related to comp code, see https://github.com/SSL-A-Team/software/issues/160


ateam_common/CMakeLists.txt
```diff
@@ -22,14 +23,13 @@ add_library(${PROJECT_NAME} SHARED
...
-  src/team_info_listener.cpp
-  src/game_state_listener.cpp
+  src/game_controller_listener.cpp
 )
-set_target_properties(${PROJECT_NAME} PROPERTIES CXX_STANDARD 17)
+set_target_properties(${PROJECT_NAME} PROPERTIES CXX_STANDARD 20)
 ament_target_dependencies(${PROJECT_NAME}
   rclcpp
   rclcpp_components
-  Eigen3
```
barulicm, line 32: Eigen is still used in this package. Can anyone remember why we removed it here?

ateam_common/src/game_controller_listener.cpp
```diff
@@ -86,13 +92,46 @@ void TeamInfoListener::RefereeMessageCallback(
...
+  // only force start and above have this mapping
+  uint8_t command = msg->command;
+  if (team_color_ != TeamColor::Unknown &&
+    command > static_cast<std::underlying_type_t<GameCommand>>(GameCommand::ForceStart))
+  {
+    if (team_color_ == TeamColor::Blue) {
+      bool command_is_blue = command % 2 == 1;
+      command += (command_is_blue ? -1 : 1);
+    }
+  }
```
barulicm, line 124: While clever, this feels brittle to changes in the command codes and feels less maintainable than simply checking the commands against the named constants and handling them.

ateam_common/src/multicast_receiver.cpp
```diff
@@ -32,16 +37,25 @@ namespace ateam_common
...
-  multicast_socket_.set_option(boost::asio::ip::multicast::join_group(multicast_address));
+  if (interface_address.empty()) {
+    multicast_socket_.set_option(boost::asio::ip::multicast::join_group(multicast_address));
+  } else {
+    multicast_socket_.set_option(
+      boost::asio::ip::multicast::join_group(
+        multicast_address,
+        boost::asio::ip::make_address_v4(interface_address)));
+  }
+  // JoinMulticastGroupOnAllV4Interfaces(multicast_address);
```
barulicm, line 58: We should remove this (and the function definitions below) since I never got it working at comp. Can come back around and try it again if we want to in the future.

```suggestion
```

ateam_common/test/test_robot_assignment.cpp
```diff
@@ -0,0 +1,108 @@
...
+    {0, 0},
+    {1, 1},
+    {2, 2},
+    {3, 3},
+  }};
+  auto result1 = assign(robots1, goals1);
+  EXPECT_EQ(result1, exp_result1);
+
+
+  // less goals than robots
```
barulicm, line 54: We should break these out into different, named tests instead of having them just be commented groups in one test.

ateam_common/src/multicast_receiver.cpp
fourpenny: Can you mark the other functions here that need to be removed/edit this file? I'm not sure if you mean everything below line 58 here or something else based on your comment here.

ateam_common/test/test_robot_assignment.cpp
fourpenny: Done

ateam_common/CMakeLists.txt
fourpenny: I added this back in... on line 13, do we need the `eigen3_cmake_module`? When looking at the package here on GitHub https://github.com/ros2/eigen3_cmake_module, it looks like this is to address issues with Ubuntu Bionic (18.04 LTS) which I assume none of us are using with ROS2 Humble, since it's not compatible.

ateam_common/src/game_controller_listener.cpp
fourpenny: I replaced this with a switch statement that should have the same effect (translating the format from `ssl_league_msgs` into our `ateam_common` `GameCommand` type). Can you double check that the logic looks okay and that the code is more understandable?

ateam_common/CMakeLists.txt
barulicm: Does a clean build work without it?
I know the readme mentions Bionic, but I don't think the Eigen setup has changed in recent releases, so it might still be relevant.

ateam_common/test/test_robot_assignment.cpp
barulicm: I see you've added new tests, but those new ones don't actually test anything. It looks like they're only setting up the inputs and expected results.

ateam_geometry/include/ateam_geometry/ateam_geometry.hpp
barulicm, line 1: Any ideas on why the ateam_geometry changes are showing up in this diff?

ateam_common/src/game_controller_listener.cpp
```diff
@@ -103,26 +103,86 @@ void GameControllerListener::RefereeMessageCallback(
...
-      bool command_is_blue = command % 2 == 1;
-      command += (command_is_blue ? -1 : 1);
-    }
+
+  // Cases in the switch here correspond with the options in the Referee message
+  // definition in the ssl_league_msgs package, translated to be for our/their
+  // team based on whether we are currently TeamColor::Yellow or TeamColor::Blue
+  // to the GameCommand format defined in the GameControllerListener's header file.
+  switch (command) {
+    case 0:
```
barulicm, line 113: We should use the named constants from [the message definition](https://github.com/SSL-A-Team/software/blob/main/ssl_league_msgs/msg/game_controller/Referee.msg)

For example:
```
    case ssl_league_msgs::msg::Referee::COMMAND_HALT:
```

ateam_geometry/include/ateam_geometry/ateam_geometry.hpp
fourpenny: I did a squash commit here, which I think may have resulted in these files looking modified rather than simply updating. I'll look into whether I can fix this if it's important to keep the Git history accurate for these comp updates.

ateam_common/CMakeLists.txt
fourpenny: A clean build with line 13 commented out passes our tests and successfully launches the simulation bringup file. But maybe not worth removing, just to be safe.

ateam_common/test/test_robot_assignment.cpp
fourpenny: Whoops, I can't read. Added the `EXPECT_EQ` macros here that should actually implement real tests now.

ateam_geometry/include/ateam_geometry/ateam_geometry.hpp
fourpenny: Looks like a proper merge fixed this issue. Resolving.

ateam_common/src/game_controller_listener.cpp
fourpenny: Done

ateam_common/src/multicast_receiver.cpp
fourpenny: Removed this and the corresponding function definition.

ateam_common/CMakeLists.txt
barulicm: Fair enough. Good to know though.

ateam_common/src/multicast_receiver.cpp
fourpenny: Done!

barulicm (approved):

PR #167, opened 2023-10-29
Dont use tracks if they have actually zero validity score

CollinAvidano:


fourpenny (approved):
looks good to me :)

fourpenny:
Let's add this as part of the comp vision merge

PR #170, opened 2023-10-30
Comp 2023 - ateam_kenobi

fourpenny:
Brings in all the code from Kenobi with no edits...
Let's work together to review this for style/basic logic. If you notice or remember bugs as you go through things, add them to the repo issues. Feel free to group multiple minor bugs together as appropriate/add TODOs.

### DIFF
--- a/ateam_kenobi/src/in_play_eval.hpp
+++ b/ateam_kenobi/src/in_play_eval.hpp
@@ -1,0 +1,131 @@
+// Copyright 2021 A Team
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL
+// THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+
+#ifndef IN_PLAY_EVAL_HPP_
+#define IN_PLAY_EVAL_HPP_
+
+#include <math.h>
+#include <CGAL/squared_distance_2.h>
+#include <Eigen/Dense>
+#include <chrono>
+#include <optional>
+#include "types/world.hpp"
+#include <ateam_geometry/types.hpp>
+#include <ateam_common/game_controller_listener.hpp>
+
+
+/**
+ *  Small class to be "ticked" with the world output_state and keep track of if we enter a kickoff and then when we exit (ball has moved 0.05 after normal_start is issued, 10 seconds pass, force start is issued)
+ *
+*/
+class InPlayEval
+{
+public:
+  // Ctors
+  InPlayEval() {}
+
+  // Vars
+  bool in_play {};
+  bool our_penalty {};
+  bool their_penalty {};
+  ateam_common::GameCommand cur_state {ateam_common::GameCommand::Stop};
+  // optional in case ball cant be seen
+  std::optional<ateam_geometry::Point> maybe_kickoff_position {};
+
+  static constexpr double DIST_THRESHOLD {0.05};
+
+  // only 3 things enter us into play according to the appendix B state machine
+  void update(ateam_kenobi::World & world)
+  {
+    // edge triggered
+    // we just need an edge on commands to know not to do in_play = false in the same normal_play
+    ateam_common::GameCommand next_command = world.referee_info.running_command;
+    if (cur_state != next_command) {
+      switch (next_command) {
+        // all of these are not in play technically but not for this purpose
+        case ateam_common::GameCommand::Halt:
+        case ateam_common::GameCommand::Stop:
+        case ateam_common::GameCommand::TimeoutTheirs:
+
+        case ateam_common::GameCommand::PrepareKickoffOurs:
+        case ateam_common::GameCommand::PrepareKickoffTheirs:
+        case ateam_common::GameCommand::DirectFreeOurs:
+        case ateam_common::GameCommand::DirectFreeTheirs:
+        case ateam_common::GameCommand::BallPlacementOurs:
+        case ateam_common::GameCommand::BallPlacementTheirs:
+        case ateam_common::GameCommand::NormalStart:
+          // reset in_play in any of these state
+          world.in_play = false;
+          break;
+
+        case ateam_common::GameCommand::PreparePenaltyOurs:
+          world.in_play = false;
+          world.our_penalty = true;
+          world.their_penalty = false;
+          break;
+        case ateam_common::GameCommand::PreparePenaltyTheirs:
+          world.in_play = false;
+          world.their_penalty = true;
+          world.our_penalty = false;
+          break;
+
+        // force play really play acts the same as play
+        case ateam_common::GameCommand::ForceStart:
+          world.in_play = true;
+          world.their_penalty = false;
+          world.our_penalty = false;
+          break;
+        default:
+          world.in_play = true;
+          world.their_penalty = false;
+          world.our_penalty = false;
+      }
+    }
+
+    cur_state = next_command;
+
+    // level
+    if (in_play == false) {
+      auto ball_current = world.ball;
+      if (maybe_kickoff_position.has_value()) {
+        // if we are tracking a start position and a current ball and its further away than
+        // dist we are in play
+        if (sqrt(CGAL::squared_distance(ball_current.pos, maybe_kickoff_position.value())) >
+          DIST_THRESHOLD)
+        {
+          world.in_play = true;
+          // so if you read this and its any of the states in the above you are basically
+          // playing under restrictions
+        }
+      } else {
+        // set the ball position if we dont have one yet during this kickoff
+        maybe_kickoff_position = ball_current.pos;
+      }
+    } else {
+      // note we may not have a view of the ball the frame we enter kickoff thats
+      // why this is seperate
+      maybe_kickoff_position = std::nullopt;
+    }
+    return;
+  }
+};
+
+#endif  // IN_PLAY_EVAL_HPP_
--- a/ateam_kenobi/src/kenobi_node.cpp
+++ b/ateam_kenobi/src/kenobi_node.cpp
@@ -1,0 +1,266 @@
+// Copyright 2021 A Team
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL
+// THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+
+#include <tf2/convert.h>
+#include <tf2/utils.h>
+#include <chrono>
+#include <functional>
+#include <rclcpp/rclcpp.hpp>
+#include <rclcpp_components/register_node_macro.hpp>
+#include <ateam_msgs/msg/ball_state.hpp>
+#include <ateam_msgs/msg/robot_state.hpp>
+#include <ateam_msgs/msg/field_info.hpp>
+#include <ateam_msgs/msg/robot_motion_command.hpp>
+#include <ateam_msgs/msg/overlay.hpp>
+#include <ateam_msgs/msg/world.hpp>
+#include <ateam_common/game_controller_listener.hpp>
+#include <ateam_common/topic_names.hpp>
+#include <ateam_common/indexed_topic_helpers.hpp>
+#include <ateam_geometry/types.hpp>
+#include <tf2_geometry_msgs/tf2_geometry_msgs.hpp>
+#include "types/world.hpp"
+#include "types/message_conversions.hpp"
+#include "play_selector.hpp"
+#include "visualization/overlay_publisher.hpp"
+#include "in_play_eval.hpp"
+#include "motion/world_to_body_vel.hpp"
+
+namespace ateam_kenobi
+{
+
+using ateam_common::indexed_topic_helpers::create_indexed_publishers;
+using ateam_common::indexed_topic_helpers::create_indexed_subscribers;
+using namespace std::literals::chrono_literals;
+
+class KenobiNode : public rclcpp::Node
+{
+public:
+  explicit KenobiNode(const rclcpp::NodeOptions & options = rclcpp::NodeOptions())
+  : rclcpp::Node("kenobi_node", options),
+    overlay_publisher_("Kenobi", *this),
+    play_info_publisher_(*this),
+    play_selector_(overlay_publisher_, play_info_publisher_),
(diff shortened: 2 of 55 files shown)


ateam_kenobi/src/in_play_eval.hpp
```diff
@@ -0,0 +1,131 @@
...
+#include <Eigen/Dense>
+#include <chrono>
+#include <optional>
+#include "types/world.hpp"
+#include <ateam_geometry/types.hpp>
+#include <ateam_common/game_controller_listener.hpp>
+
+
+/**
+ *  Small class to be "ticked" with the world output_state and keep track of if we enter a kickoff and then when we exit (ball has moved 0.05 after normal_start is issued, 10 seconds pass, force start is issued)
```
fourpenny, line 36: It's unclear from the rules and I don't quite remember from comp... Is a force start automatically sent 10 seconds after kickoff if the ball doesn't move 0.05m? We don't seem to keep track of time this class, so I'm trying to figure out whether we actually need to.

ateam_kenobi/src/in_play_eval.hpp
```diff
@@ -0,0 +1,131 @@
...
+          world.in_play = true;
+          // so if you read this and its any of the states in the above you are basically
+          // playing under restrictions
+        }
+      } else {
+        // set the ball position if we dont have one yet during this kickoff
+        maybe_kickoff_position = ball_current.pos;
+      }
+    } else {
+      // note we may not have a view of the ball the frame we enter kickoff thats
```
fourpenny, line 123: Below resets the kick off position before the next instance where we go out of play? Is that the actual purpose?

ateam_kenobi/src/motion/motion_controller.cpp
```diff
@@ -0,0 +1,195 @@
...
+// PID gains
+CREATE_PARAM(double, "motion/pid/x_kp", x_kp, 2);
+CREATE_PARAM(double, "motion/pid/y_kp", y_kp, 2);
+CREATE_PARAM(double, "motion/pid/t_kp", t_kp, 3);
+
+// PID velocity limits
+CREATE_PARAM(double, "motion/pid/x_max", x_max, 2);
+CREATE_PARAM(double, "motion/pid/y_max", y_max, 2);
+CREATE_PARAM(double, "motion/pid/t_max", t_max, 4);
+*/
```
fourpenny, line 44: Do we still need these commented out PID values?

ateam_kenobi/src/kenobi_node.cpp
```diff
@@ -0,0 +1,266 @@
...
+    play_info_publisher_(*this),
+    play_selector_(overlay_publisher_, play_info_publisher_),
+    game_controller_listener_(*this)
+  {
+    declare_parameter<bool>("use_world_velocities", false);
+
+    create_indexed_subscribers<ateam_msgs::msg::RobotState>(
+      blue_robots_subscriptions_,
+      Topics::kBlueTeamRobotPrefix,
+      10,
```
fourpenny, line 68: A bunch of these subscribers use the same value for the QOS, do we want to make this a constant/variable that they share?

ateam_kenobi/src/kenobi_node.cpp
```diff
@@ -0,0 +1,266 @@
...
+
+  // TODO(CAVIDANO): REMOVE THE THEIR ROBOTS HERE THIS SHOULD NOT ASSIGN ANYTHING IN THE NOT CASE
+  // AS IT ALSO CATCHES UNKOWN TEAM
+  void yellow_robot_state_callback(
+    const ateam_msgs::msg::RobotState::SharedPtr robot_state_msg,
+    int id)
+  {
+    const auto are_we_yellow = game_controller_listener_.GetTeamColor() ==
+      ateam_common::TeamColor::Yellow;
+    auto & robot_state_array = are_we_yellow ? world_.our_robots : world_.their_robots;
```
fourpenny, line 142: Change at some point for lines 130 and 142 to address case of Unknown team color (following Collin's TODO)

ateam_kenobi/src/kenobi_node.cpp
```diff
@@ -0,0 +1,266 @@
...
+    robot_state_callback(robot_state_array, id, robot_state_msg);
+  }
+
+  void robot_state_callback(
```
fourpenny, line 146: Maybe we should document what some of these callbacks do? They're pretty easy to read but would make it faster to find things going through this file

ateam_kenobi/src/path_planning/path_planner.cpp
```diff
@@ -0,0 +1,233 @@
...
+  auto their_robot_obstacles = world.their_robots |
+    std::views::filter(std::mem_fn(&std::optional<Robot>::has_value)) |
+    std::views::transform(obstacle_from_robot);
+  obstacles.insert(
+    obstacles.end(),
+    their_robot_obstacles.begin(),
+    their_robot_obstacles.end());
+}
+
+void PathPlanner::addDefaultObstacles(
```
fourpenny, line 211: At some point, we want to move some of the common field geometries (goal box, center line, etc) to a separate file in ateam_common or something to make it easier to reference them

ateam_kenobi/src/path_planning/path_planner.cpp
```diff
@@ -0,0 +1,233 @@
...
+    addDefaultObstacles(world, augmented_obstacles);
+  }
+
+  if (options.avoid_ball) {
+    augmented_obstacles.push_back(ateam_geometry::makeCircle(world.ball.pos, 0.04267 / 2));
+  }
+
+  // if (!isStateValid(start, world, augmented_obstacles, options)) {
+  //   return {};
+  // }
```
fourpenny, line 57: Remove duplicate commented code

ateam_kenobi/src/path_planning/path_planner.cpp
```diff
@@ -0,0 +1,233 @@
...
+  std::vector<ateam_geometry::AnyShape> augmented_obstacles = obstacles;
+
+  addRobotsToObstacles(world, start, augmented_obstacles);
+
+  if (options.use_default_obstacles) {
+    addDefaultObstacles(world, augmented_obstacles);
+  }
+
+  if (options.avoid_ball) {
+    augmented_obstacles.push_back(ateam_geometry::makeCircle(world.ball.pos, 0.04267 / 2));
```
fourpenny, line 52: Why this number

ateam_kenobi/src/path_planning/path_planner.cpp
```diff
@@ -0,0 +1,233 @@
...
+  const ateam_geometry::Point & state,
+  const World & world,
+  const std::vector<ateam_geometry::AnyShape> & obstacles,
+  const PlannerOptions & options)
+{
+  const auto x_bound = (world.field.field_length / 2.0) + world.field.boundary_width - kRobotRadius;
+  const auto y_bound = (world.field.field_width / 2.0) + world.field.boundary_width - kRobotRadius;
+  const ateam_geometry::Rectangle pathable_region(
+    ateam_geometry::Point(-x_bound, -y_bound),
+    ateam_geometry::Point(x_bound, y_bound));
```
fourpenny, line 103: We end up calling this function a lot, and this value doesn't change. Might be better to pull this out as a class member instead of recalculating it every time we want to see if a point is valid.

ateam_kenobi/src/types/field.hpp
```diff
@@ -0,0 +1,53 @@
...
+struct Field
+{
+  // we will definetly change the format of this at some point this is preliminary
+  // since we dont really have a geometry library yet
+  // all of this should be in our coordinate system
+  // There is a mirroring ros message of this type
+  float field_length;
+  float field_width;
+  // NOTE GOAL DEPTH AND WIDTH ARE THE SIZE OF THE GOALIE BOX NOT OF THE GOAL ITSELF
+  // TODO(CAVIDANO) all values in here are currently inaccurate
```
fourpenny, line 43: @CollinAvidano are the values here actually inaccurate like the TODO says

ateam_kenobi/src/types/message_conversions.cpp
```diff
@@ -0,0 +1,127 @@
...
+  world_msg.field = toMsg(obj.field);
+  world_msg.referee_info = toMsg(obj.referee_info);
+
+  world_msg.balls.push_back(toMsg(obj.ball));
+
+  for (const auto & maybe_robot : obj.our_robots) {
+    if (maybe_robot.has_value()) {
+      world_msg.our_robots.push_back(toMsg(maybe_robot.value()));
+    } else {
+      world_msg.our_robots.push_back(ateam_msgs::msg::RobotState());
```
fourpenny, line 112: What's the initialization state of the RobotStateMsg?

ateam_kenobi/src/types/world.hpp
```diff
@@ -0,0 +1,54 @@
...
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+
+#ifndef TYPES__WORLD_HPP_
+#define TYPES__WORLD_HPP_
+
+#include <optional>
+#include <array>
+#include <vector>
```
fourpenny, line 27: Don't think we need vector anymore here

ateam_kenobi/src/play_helpers/available_robots.cpp
```diff
@@ -0,0 +1,66 @@
...
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+
+#include "available_robots.hpp"
+
+namespace ateam_kenobi::play_helpers
+{
+
+std::vector<Robot> getAvailableRobots(const World & world)
```
fourpenny, line 27: TODO? Add remove goalie function here automatically unless there's some reason we wouldn't do that? We seem to use it in each of the plays anyway

ateam_kenobi/src/play_selector.cpp
```diff
@@ -0,0 +1,166 @@
...
+    return finalizeSelection(&our_penalty_play_, current_game_command);
+  }
+
+  if (world.their_penalty) {
+    return finalizeSelection(&their_penalty_play_, current_game_command);
+  }
+
+  // if(world.in_play) {
+  //   return finalizeSelection(&basic_122_play_, current_game_command);
+  // }
```
fourpenny, line 66: Is there a reason this is commented out?

fourpenny: We need to revisit/revise plays in general so it doesn't really matter

ateam_kenobi/src/play_selector.cpp
```diff
@@ -0,0 +1,166 @@
...
+
+  plays::BasePlay * selected_play = &halt_play_;
+
+  switch (current_game_command) {
+    case ateam_common::GameCommand::Halt:
+    case ateam_common::GameCommand::TimeoutOurs:
+    case ateam_common::GameCommand::TimeoutTheirs:
+    case ateam_common::GameCommand::GoalOurs:
+    case ateam_common::GameCommand::GoalTheirs:
```
fourpenny, line 75: Repeat of logic on lines 48-54, we probably just need one

ateam_kenobi/src/plays/basic_122.hpp
```diff
@@ -0,0 +1,61 @@
...
+
+private:
+  skills::LineKick striker_skill_;
+  skills::Blockers blockers_skill_;
+  skills::Goalie goalie_skill_;
+
+  void assignAndRunStriker(
+    std::vector<Robot> & available_robots, const World & world,
+    std::array<std::optional<ateam_msgs::msg::RobotMotionCommand>,
+    16> & motino_commands);
```
fourpenny, line 51: This is misspelled here and on 56

ateam_kenobi/src/plays/stop_play.cpp
```diff
@@ -0,0 +1,87 @@
...
+
+std::array<std::optional<ateam_msgs::msg::RobotMotionCommand>, 16> StopPlay::runFrame(
+  const World & world)
+{
+  std::array<std::optional<ateam_msgs::msg::RobotMotionCommand>, 16> stop_motion_commands;
+  for (size_t robot_id = 0; robot_id < 16; ++robot_id) {
+    // only going to do 8 robots for now to avoid crowding on one side of the ball
+    // avoid the ball by 0.7m just to be safe
+    double radius = 0.7;
+    const auto & maybe_robot = world.our_robots[robot_id];
```
fourpenny, line 51: This play needs to be revised for sure

ateam_kenobi/src/plays/stop_play.cpp
```diff
@@ -0,0 +1,87 @@
...
+  }
+  // Draw Keepout Circle
+  overlay_publisher_.drawCircle(
+    "keepout_circle",
+    ateam_geometry::makeCircle(world.ball.pos, 0.35), "red", "transparent");
+
+  // overlay_publisher_.drawSegment(
+  //   "keepout_circle",
+  //   ateam_geometry::makeCircle(world.ball.pos, 0.35), "red", "transparent");
+
```
fourpenny, line 82: Remove extra comments

ateam_kenobi/src/skills/line_kick.cpp
```diff
@@ -0,0 +1,120 @@
...
+  };
+  return easy_move_to_.runFrame(robot, world, obstacles);
+}
+
+ateam_msgs::msg::RobotMotionCommand LineKick::faceBall(const World & world, const Robot & robot)
+{
+  easy_move_to_.setPlannerOptions({});
+  easy_move_to_.setTargetPosition(robot.pos);
+  easy_move_to_.face_point(target_point_);
+  easy_move_to_.setMaxAngularVelocity(1.0);
```
fourpenny, line 97: Curious, do we ever change this back to the default? Does it change automatically?

fourpenny: This affects MotionController btw

fourpenny: In general we need to check whether we reset the motion planner options/settings or reinitialize those objects later

ateam_kenobi/src/plays/halt_play.hpp
```diff
@@ -0,0 +1,49 @@
...
+
+  void reset() override;
+
+  std::array<std::optional<ateam_msgs::msg::RobotMotionCommand>,
+    16> runFrame(const World & world) override;
+
+private:
+  path_planning::PathPlanner path_planner_;
+  MotionController motion_controller_;
+  int prev_assigned_id_ = -1;
```
fourpenny, line 45: I don't think we need this

ateam_kenobi/src/plays/normal_play.hpp
fourpenny, line 1: We don't use normal play but maybe this will be useful later?

ateam_kenobi/src/plays/test_play.hpp
```diff
@@ -0,0 +1,48 @@
...
+#include "skills/goalie.hpp"
+
+namespace ateam_kenobi::plays
+{
+class TestPlay : public BasePlay
+{
+public:
+  explicit TestPlay(
+    visualization::OverlayPublisher & overlay_publisher,
+    visualization::PlayInfoPublisher & play_info_publisher);
```
fourpenny, line 36: Do we care if these have a consistent naming style?

ateam_kenobi/src/plays/wall_play.cpp
```diff
@@ -0,0 +1,134 @@
...
+      target.x(),
+      target.y() + (spacing.y() * i)
+      // both of these were source in the last nothing else changed
+    );
+
+    points_on_segment.push_back(segment_point);
+    // WHAT THE ACTUAL FUCK
+  }
+  return points_on_segment;
+}
```
fourpenny, line 61: I think we could replace this with the logic that's used in the penalty and stop plays to form a pattern instead

fourpenny: CGAL is hard :(

ateam_kenobi/src/skills/kick.hpp
fourpenny, line 1: I think we only use this in kickoff and probably should replace it with line kick

ateam_kenobi/src/robot_assignment.hpp
```diff
@@ -0,0 +1,93 @@
...
+
+  /*
+   * Hacky version that just assigns robots in order
+   */
+  std::unordered_map<size_t, size_t> assignments;
+  auto num_assignments = std::min(available_robots.size(), goal_positions.size());
+  for (auto index = 0ul; index < num_assignments; ++index) {
+    assignments[available_robots[index].id] = index;
+  }
+  return assignments;
```
fourpenny, line 60: We should fix assignment so we don't have to do hacky assignment

ateam_kenobi/src/plays/basic_122.cpp
fourpenny, line 1: Right now the back 2 robots don't do anything in basic 1-2-2, do we want them to do something...?

ateam_kenobi/src/in_play_eval.hpp
barulicm: I don't think the rules specify that a force start happens automatically, but I think in practice a force start is called. Probably don't need to worry about this for merging.

ateam_kenobi/src/in_play_eval.hpp
barulicm: This seems to reset the tracked kickoff position when the ball is in play so that the next time we start tracking whether the ball's in play we don't accidentally carry over state from this time. Seems reasonable.

ateam_kenobi/src/motion/motion_controller.cpp
barulicm: probably not

ateam_kenobi/src/kenobi_node.cpp
barulicm: meh 🤷 we can, but it's also not that bad to not

ateam_kenobi/src/path_planning/path_planner.cpp
barulicm: That is the diameter of a golf ball in meters.

ateam_kenobi/src/path_planning/path_planner.cpp
barulicm: Just need to be careful about the "this doesn't change part" because I believe this comes from vision, so it will probably change at least once when we start getting field geometry packets.

ateam_kenobi/src/play_helpers/available_robots.cpp
barulicm: Hmmmmm...... I think I'd like to keep this function, but I wouldn't be opposed to adding something like a `getNonGoalieRobots()` as a convenience

barulicm (approved):
We'll fix it live

PR #182, opened 2023-12-06
Add Control Debug Telemetry Data Pipeline

guyfleeman:
This is the software side of the initial prototype to get motion control debug data into the ROS ecosystem in real time for logging and analysis purposes.

### DIFF
--- a/ateam_msgs/CMakeLists.txt
+++ b/ateam_msgs/CMakeLists.txt
@@ -6,6 +6,7 @@ find_package(rosidl_default_generators REQUIRED)
 find_package(builtin_interfaces REQUIRED)
 find_package(std_msgs REQUIRED)
 find_package(geometry_msgs REQUIRED)
+find_package(sensor_msgs REQUIRED)
 find_package(ssl_league_msgs REQUIRED)
 
 rosidl_generate_interfaces(${PROJECT_NAME}
@@ -44,6 +45,7 @@ rosidl_generate_interfaces(${PROJECT_NAME}
   builtin_interfaces
   std_msgs
   geometry_msgs
+  sensor_msgs
   ssl_league_msgs
   ADD_LINTER_TESTS
 )
--- a/ateam_msgs/CMakeLists.txt
+++ b/ateam_msgs/CMakeLists.txt
@@ -17,6 +17,8 @@ rosidl_generate_interfaces(${PROJECT_NAME}
   msg/RefereeInfo.msg
   msg/RobotFeedback.msg
   msg/RobotMotionCommand.msg
+  msg/RobotMotionFeedback.msg
+  msg/RobotMotorFeedback.msg
   msg/RobotState.msg
   msg/Sample3d.msg
   msg/TeamClientConnectionStatus.msg
--- a/ateam_radio_bridge/src/conversion.cpp
+++ b/ateam_radio_bridge/src/conversion.cpp
@@ -64,6 +64,20 @@ ateam_msgs::msg::RobotFeedback ConvertBasicTelemetry(const BasicTelemetry & basi
   return robot_feedback;
 }
 
+void ConvertFloatArrayToVec3(const float (&float_arr_3)[3], geometry_msgs::msg::Vector3 & vec) {
+  vec.x = float_arr_3[0];
+  vec.y = float_arr_3[1];
+  vec.z = float_arr_3[2];
+}
+
+geometry_msgs::msg::Vector3 ConvertFloatArrayToVec3(const float (&float_arr_3)[3]) {
+  geometry_msgs::msg::Vector3 vec;
+
+  ConvertFloatArrayToVec3(float_arr_3, vec);
+
+  return vec;
+}
+
 void ConvertFloatArrayToTwist(const float (&fw_state_space_array)[3], geometry_msgs::msg::Twist & twist) {
   twist.linear.x = fw_state_space_array[0];
   twist.linear.y = fw_state_space_array[1];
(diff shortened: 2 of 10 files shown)


guyfleeman:
This obviously needs testing on hardware, then we should have a discussion about meaningful things like code quality and then probably a discussion about names. I know the software-communications packets have different names for stuff than the ROS messages. Software and firmware motion nomenclature was already conflicted before these additions, so perhaps we should merge this as is and then do a big cleanup of packet structure and names.

No action needed at this time.

guyfleeman:
added wheel vels, and imu. Needs testing. Especially look at wheel vels, firmware and software do not use the same ordering for wheels in float[4] arrays representing wheel state

ateam_radio_bridge/src/conversion.hpp
```diff
@@ -24,15 +24,17 @@
...
+#include <control_debug_telemetry.h>
 #include <hello_data.h>
 
 namespace ateam_radio_bridge
 {
 
-ateam_msgs::msg::RobotFeedback Convert(const BasicTelemetry & basic_telemetry);
-
+ateam_msgs::msg::RobotFeedback ConvertBasicTelemetry(const BasicTelemetry & basic_telemetry);
+ateam_msgs::msg::RobotMotionFeedback ConvertControlDebugTelemetry(const ControlDebugTelemetry & control_debug_telemetry);
```
barulicm, line 37: nit: most of our conversion helpers in other packages stick to a consistent name and let overload deduction pick the right one for us.

So this section could end up looking like this:
```suggestion
ateam_msgs::msg::RobotFeedback Convert(const BasicTelemetry & basic_telemetry);
ateam_msgs::msg::RobotMotionFeedback Convert(const ControlDebugTelemetry & control_debug_telemetry);
```

ateam_radio_bridge/src/conversion.cpp
```diff
@@ -64,4 +64,75 @@ ateam_msgs::msg::RobotFeedback Convert(const BasicTelemetry & basic_telemetry)
...
+  robot_motion_feedback.body_velocity_state_estimate = ConvertFloatArrayToTwist(control_debug_telemetry.cgkf_body_velocity_state_estimate);
+  robot_motion_feedback.body_velocity_control_variable = ConvertFloatArrayToTwist(control_debug_telemetry.body_velocity_u);
+
+  // copying from some C arrays here, not a struct so add some sanity checks
+  assert((sizeof(control_debug_telemetry.wheel_velocity_u) / sizeof(control_debug_telemetry.wheel_velocity_u[0]) <= robot_motion_feedback.wheel_velocity_control_variable.size()));
+  assert((sizeof(control_debug_telemetry.wheel_velocity_clamped_u) / sizeof(control_debug_telemetry.wheel_velocity_clamped_u[0]) <= robot_motion_feedback.clamped_wheel_velocity_control_variable.size()));
+  std::copy(std::begin(control_debug_telemetry.wheel_velocity_u), std::end(control_debug_telemetry.wheel_velocity_u), std::begin(robot_motion_feedback.wheel_velocity_control_variable));
+  std::copy(std::begin(control_debug_telemetry.wheel_velocity_clamped_u), std::end(control_debug_telemetry.wheel_velocity_clamped_u), std::begin(robot_motion_feedback.clamped_wheel_velocity_control_variable));
```
barulicm, line 133: A more modern approach to this kind of thing uses [std::span](https://en.cppreference.com/w/cpp/container/span) to wrap the raw array and [std::ranges::copy](https://en.cppreference.com/w/cpp/algorithm/ranges/copy) for a shorter call to copy the whole container.

```suggestion
  std::span wheel_velocity_u_span{control_debug_telemetry.wheel_velocity_u};
  static_assert(wheel_velocity_u_span.size() == robot_motion_feedback.wheel_velocity_control_variable.size());
  std::ranges::copy(wheel_velocity_u_span, robot_motion_feedback.wheel_velocity_control_variable.begin());

  std::span wheel_velocity_clamped_u_span{control_debug_telemetry.wheel_velocity_clamped_u};
  static_assert(wheel_velocity_clamped_u_span.size() == robot_motion_feedback.clamped_wheel_velocity_control_variable.size());
  std::ranges::copy(wheel_velocity_clamped_u_span, robot_motion_feedback.clamped_wheel_velocity_control_variable.begin());
```

ateam_radio_bridge/src/conversion.cpp
```diff
@@ -64,4 +64,75 @@ ateam_msgs::msg::RobotFeedback Convert(const BasicTelemetry & basic_telemetry)
...
+  twist.angular.z = fw_state_space_array[2];
+}
+
+geometry_msgs::msg::Twist ConvertFloatArrayToTwist(const float (&fw_state_space_array)[3]) {
+  geometry_msgs::msg::Twist twist;
+
+  ConvertFloatArrayToTwist(fw_state_space_array, twist);
+
+  return twist;
+}
```
barulicm, line 96: What's the motivation for the extra internal helper that returns void rather than populating the twist directly in the function called by the top-level conversion?

ateam_radio_bridge/src/conversion.cpp
```diff
@@ -64,4 +64,75 @@ ateam_msgs::msg::RobotFeedback Convert(const BasicTelemetry & basic_telemetry)
...
+
+ateam_msgs::msg::RobotMotionFeedback ConvertControlDebugTelemetry(const ControlDebugTelemetry & control_debug_telemetry) {
+  ateam_msgs::msg::RobotMotionFeedback robot_motion_feedback;
+
+  robot_motion_feedback.motors[robot_motion_feedback.FRONT_LEFT_MOTOR] = ConvertMotorDebugTelemetry(control_debug_telemetry.motor_fl);
+  robot_motion_feedback.motors[robot_motion_feedback.FRONT_RIGHT_MOTOR] = ConvertMotorDebugTelemetry(control_debug_telemetry.motor_fr);
+  robot_motion_feedback.motors[robot_motion_feedback.BACK_RIGHT_MOTOR] = ConvertMotorDebugTelemetry(control_debug_telemetry.motor_br);
+  robot_motion_feedback.motors[robot_motion_feedback.BACK_LEFT_MOTOR] = ConvertMotorDebugTelemetry(control_debug_telemetry.motor_bl);
+
+  robot_motion_feedback.imu.orientation_covariance[0] = -1.0;  // ROS2 docs say if a sensor doesn't provide a data point, then set element '0' of it's covariance to
```
barulicm, line 120: incomplete comment?

```suggestion
  robot_motion_feedback.imu.orientation_covariance[0] = -1.0;  // ROS2 docs say if a sensor doesn't provide a data point, then set element '0' of it's covariance to -1
```

ateam_radio_bridge/src/conversion.cpp
guyfleeman: probably just thinking too much like C. I guess if all the assignment operators are overloaded there's no reason to have it. I can inline it.

ateam_msgs/CMakeLists.txt
```diff
@@ -6,6 +6,7 @@ find_package(rosidl_default_generators REQUIRED)
 find_package(builtin_interfaces REQUIRED)
 find_package(std_msgs REQUIRED)
 find_package(geometry_msgs REQUIRED)
+find_package(sensor_msgs REQUIRED)
```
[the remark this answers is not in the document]
guyfleeman: should be fixed in latest commit. Why did this pass CI? it is because some other node package already flagged it as needed, so rosdep still installed it?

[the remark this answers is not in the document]
barulicm: Yeah, for a package as common as sensor_msgs, it'd be unlikely to cause a real problem in most systems.
But, if all of ROS was being built from source with this package included, for example, the dependency would be missed so colcon might try to build this package before sensor_msgs was built.

barulicm (approved):

PR #185, opened 2023-12-27
Treat all circles as solid

fourpenny:
Adds custom intersection code for circles to our ateam_geometry package to consistently treat circles as solid + tests to make sure said code works.

This hopefully will help with some of our path planner issues...


ateam_geometry/include/ateam_geometry/disk_intersection.hpp
```diff
@@ -0,0 +1,199 @@
...
+ * https://en.wikipedia.org/wiki/Disk_(mathematics)
+*/
+
+#ifndef ATEAM_GEOMETRY__DISK_INTERSECTION_HPP_
+#define ATEAM_GEOMETRY__DISK_INTERSECTION_HPP_
+
+#include <CGAL/Distance_2/Point_2_Segment_2.h>
+#include <variant>
+#include <algorithm>
+#include <iostream>
```
fourpenny, line 35: TODO: remove iostream, was just here for testing

ateam_radio_bridge/software-communication
fourpenny, line 1: Can someone make sure this isn't merging in things it shouldn't? I ran `git submodule update --recursive --remote` on our repo but am not sure it worked properly.

ateam_geometry/include/ateam_geometry/variant_do_intersect.hpp
```diff
@@ -43,10 +44,31 @@ namespace ateam_geometry
...
+      } else if constexpr (std::is_same_v<ObjA, Point>) {
+        return diskPointIntersection(std::get<Circle>(object_b), object_a);
+      }
+    }
+  }
+  bool intersects = std::visit(
     [&object_a](const auto & b) {
       return CGAL::do_intersect(object_a, b);
     }, object_b);
+  return intersects;
```
barulicm, line 71: If we change our disk intersection functions to have the same name (ie. `ateam_geometry::do_disk_intersect()`) then we can simplify this function to just use `std::visit`. That also means we won't have to edit this function every time we add a new disk intersection override.

```suggestion
  if constexpr (std::is_same_v<ObjA, Circle>) {
    return std::visit([&object_a](const auto & b) {
      return ateam_geometry::do_disk_intersect(object_a, b);
    });
  }
  if constexpr (std::is_same_v<ObjB, Circle>) {
    return std::visit([&object_b](const auto & a) {
      return ateam_geometry::do_disk_intersect(object_b, a);
    });
  }
  return std::visit([&object_a](const auto & b) {
    return CGAL::do_intersect(object_a, b);
  });
```

ateam_geometry/include/ateam_geometry/variant_do_intersect.hpp
```diff
@@ -61,15 +83,13 @@ bool variantDoIntersect(const ObjA & object_a, const AnyShape & object_b)
  * @return true The two geometry objects intersect
  * @return false The two geometry objects do not intersect
  */
-template<>
```
barulicm, line 64: What motivated removing this?

ateam_geometry/test/variant_do_intersect_test.cpp
```diff
@@ -23,12 +23,37 @@
...
+
+TEST(VariantDoIntersect, CircleToSegment)
+{
+  ateam_geometry::Circle a(ateam_geometry::Point(0, 0), 1);
+  ateam_geometry::AnyShape b = ateam_geometry::Segment(
+    ateam_geometry::Point(0, 0),
+    ateam_geometry::Point(1, 0)
+  );
+  EXPECT_TRUE(ateam_geometry::variantDoIntersect(a, b));
+}
```
barulicm, line 59: We should probably have at least one test for each of our new functions and each should include a intersecting and non-intersecting example.

ateam_geometry/include/ateam_geometry/variant_do_intersect.hpp
fourpenny: I thought I added this by accident previously and didn't want to add anything extra that wasn't need... I'll add it back in!

ateam_geometry/test/variant_do_intersect_test.cpp
fourpenny: There should be a positive and negative test for every variant now

ateam_geometry/include/ateam_geometry/variant_do_intersect.hpp
fourpenny: Did this but with slightly tweaked formatting since object_a is templated and object_b is a variant

barulicm (approved):

PR #186, opened 2023-12-28
Fix radio bridge timeouts bug & unreliable test

barulicm:
`test_2_commandMessage` was failing intermittently because the timeout that disables motion and kicker was sometimes expired and sometimes not when that test was run. It should always be expired since no command messages are sent during the test. It was sometimes alive because of an erroneous line that reset the timeout on incoming messages from the robot. This PR removes that line and updates the test to reflect the disabled message the bridge should always be sending in this scenario.

### DIFF
--- a/ateam_radio_bridge/test/launch_tests/radio_discovery_test.py
+++ b/ateam_radio_bridge/test/launch_tests/radio_discovery_test.py
@@ -181,10 +181,10 @@ class TestRadioBridgeNode(unittest.TestCase):
     def test_2_commandMessage(self):
         data = self.sock.recv(508)
         expected_data = bytearray(b"\x00") * 36
-        expected_data[0:4] = [92, 255, 234, 93]  # CRC
+        expected_data[0:4] = [57, 152, 86, 229]  # CRC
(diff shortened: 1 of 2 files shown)


ateam_radio_bridge/test/launch_tests/radio_discovery_test.py
```diff
@@ -181,10 +181,10 @@ def test_1_telemetryMessage(self):
     def test_2_commandMessage(self):
         data = self.sock.recv(508)
         expected_data = bytearray(b"\x00") * 36
-        expected_data[0:4] = [92, 255, 234, 93]  # CRC
+        expected_data[0:4] = [57, 152, 86, 229]  # CRC
```
fourpenny, line 184: Why did we change the CRC here?

fourpenny (approved):
Changes make sense to me

ateam_radio_bridge/test/launch_tests/radio_discovery_test.py
barulicm: The CRC is calculated from the packet content. Since that's changed, the CRC had to change.

PR #188, opened 2024-01-13
Vision Filter Patch

barulicm:
This PR:
- Fixes the definitions of dt3
- Adjusts robot motion model covariance

These two adjustments dramatically improved the accuracy of the filter output in our lab testing.


guyfleeman (approved):
Very important, but trivial change in terms of code volume. This was run on physical robots Jan 07 and vastly improved control behavior.

PR #190, opened 2024-01-14
Fixes math for default planning obstacles.

barulicm:
Math is hard.

math

Previous bad math was putting the default obstacles for the defense areas in the wrong place and making them the wrong size. This fixes that.


chachmu (approved):
LGTM 👍

guyfleeman (approved):
woodward says I should approve this

PR #194, opened 2024-01-17
Remove extra kick vel line in radio bridge.

barulicm:
kick_vel is set to 0.0 on line 177 and then set to the actual commanded value on line 180. This PR removes line 177.


guyfleeman (approved):
woodward can't beat me to this one. Approved!

PR #195, opened 2024-01-18
Fixes bugs in UI goalie ID selector.

barulicm:
The goalie ID drop down in the UI was not working. I found two bugs to fix.

1. The way the callback was being handled for the v-select value change event was losing the binding to the state object. So, in the state's `setGoalie` method, `this` was undefined. By adding an intermediate method in the Vue component, the method is called directly on the state object and the `this` binding is handled correcty.

1. A missing leading forward slash in the service name cause roslibjs to interpret it as a relative name, yielding "/ui/team_client_node/set_desired_keeper", which doesn't exist. Adding the leading slash fixes this.

I added some console logging to `setGoalie` for better feedback in the future. Is there a way for us to show an alert if the request fails?


chachmu (approved):
:+1: Looks good. I can probably go in and make the goalie select box turn red or something if the request fails

PR #196, opened 2024-01-19
Adds team name override.

barulicm:
To scrimmage against ourselves, we need one instance of our stack to pretend to be another team, otherwise each stack won't be able to tell which "A-Team" from the GC it's supposed to use.

`GameControllerListener` already had a parameter for overriding the team name, so this PR just finishes plumbing that into the simulation bringup files.

It's worth noting that not all team names will work, because some teams use secure connections and have signatures registered with the league (including "Test Team"). Luckily, not all teams do this, so we can connect as one of the non-secure teams. "RoboJackets" is one such team that worked in my testing.


chachmu (approved):
I'll probably need to change the UI to also use the specified team name but we can do that later

PR #197, opened 2024-01-19
Changes simulator radio bridge to use local velocities.

barulicm:
Our simulator radio bridge has been using global velocities. This had the side effect of breaking our code when running as the yellow team in sim. Our gameplay code uses our conventions for the global frame, which is inverted from the sim/vision frame when we're playing as yellow in sim. This caused the robots to drive in the opposite direction than what our gameplay code was trying to request.

Turns out the sim protocol lets you control using either wheel velocities, local velocities, or global velocities. So, I just flipped it over to using local velocities like our physical radio bridge does, and now we can play as either team without issue.


chachmu (approved):
:tada: this issue was driving me crazy, I couldn't tell if it was coming from the UI or something else but I'm glad you figured it out

PR #200, opened 2024-01-23
Fixing InPlayEval

barulicm:
I'm pretty sure the reason we were seeing Kenobi get confused about when the ball was in play this weekend was because `InPlayEval` was capturing the ball's starting position at the wrong time. So, if you entered the kickoff prep command then moved the ball more than 0.5m before issuing normal start, the ball would immediately be considered in play.

This PR rewrites `InPlayEval` to hopefully follow the rules correctly. Eventually we should add real tests, but based on my informal testing in simulation, this works correctly and doesn't get tripped up by things like the ball moving during prep commands.

### DIFF
--- a/ateam_kenobi/src/in_play_eval.cpp
+++ b/ateam_kenobi/src/in_play_eval.cpp
@@ -28,26 +28,33 @@ namespace ateam_kenobi
 
 void InPlayEval::Update(World & world)
 {
+  const auto game_command = world.referee_info.running_command;
   // Force start always puts the ball in play
-  if (world.referee_info.running_command == ateam_common::GameCommand::ForceStart) {
+  if (game_command == ateam_common::GameCommand::ForceStart) {
     in_play_ = true;
   }
   // Stop / Halt puts the ball out of play
-  else if (world.referee_info.running_command == ateam_common::GameCommand::Stop ||
-    world.referee_info.running_command == ateam_common::GameCommand::Halt)
+  else if (game_command == ateam_common::GameCommand::Stop ||
+    game_command == ateam_common::GameCommand::Halt)
   {
     in_play_ = false;
     ball_start_pos_.reset();
   }
-  // Game is resuming, establish starting conditions
-  else if (IsGameResuming(world)) {
-    ball_start_pos_ = world.ball.pos;
+  // Game is exitting Stop command, decide in-play rules based on new command
+  else if (game_command != prev_game_command_ &&
+    prev_game_command_ == ateam_common::GameCommand::Stop)
+  {
     SetTimeout(world);
     SetDistanceThreshold(world);
   }
+  // Game is resuming, establish starting conditions
+  else if (IsGameResuming(world)) {
+    ball_start_pos_ = world.ball.pos;
+    timeout_start_ = std::chrono::steady_clock::now();
+  }
   // If timeout passed, ball's in play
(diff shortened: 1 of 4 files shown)


ateam_kenobi/src/in_play_eval.cpp
```diff
@@ -0,0 +1,142 @@
...
+#include "in_play_eval.hpp"
+
+#include <CGAL/squared_distance_2.h>
+#include <cmath>
+#include <limits>
+
+namespace ateam_kenobi
+{
+
+void InPlayEval::Update(World & world)
```
fourpenny, line 30: Any specific reason we changed code style in these files to use camel case function names vs. lower case ones? (Doesn't matter but eventually, maybe we should pick a single style?)

ateam_kenobi/src/in_play_eval.cpp
```diff
@@ -0,0 +1,142 @@
...
+    SetTimeout(world);
+    SetDistanceThreshold(world);
+  } else if (IsGameResuming(world)) {
+    ball_start_pos_ = world.ball.pos;
+    timeout_start_ = std::chrono::steady_clock::now();
+  } else if (HasTimeoutExpired()) {
+    in_play_ = true;
+  } else if (HasBallMoved(world)) {
+    in_play_ = true;
+  }
```
fourpenny, line 48:
```suggestion
  } else {
      in_play_ = false;  
  }
```
In case we hit an edge case that violates all of these conditions, do you think it makes sense to set `in_play_` to false?

ateam_kenobi/src/in_play_eval.cpp
```diff
@@ -0,0 +1,142 @@
...
+bool InPlayEval::HasBallMoved(const World & world)
+{
+  if (!ball_start_pos_) {
+    return false;
+  }
+
+  return CGAL::approximate_sqrt(
+    CGAL::squared_distance(
+      world.ball.pos,
+      ball_start_pos_.value())) > ball_moved_threshold_;
```
fourpenny, line 130: Do we want any epsilon here for the allowed ball distance?

fourpenny (approved):
I'm okay with this for now and think the logic makes sense, but I think we need to make actual tests for this before comp to verify.

ateam_kenobi/src/in_play_eval.cpp
barulicm: I don't remember, but I think I was matching something else.
Agreed though, I think once upon a time we talked about using the ROS style guides, which would say lower starting camel case. [ref](https://wiki.ros.org/CppStyleGuide#Function_.2F_Methods)

ateam_kenobi/src/in_play_eval.cpp
barulicm: No, the idea with this function is that all of the cases checked are when `in_play_` needs to change. If none of the conditions are hit, we just need to keep the same value we had from the last frame.

ateam_kenobi/src/in_play_eval.cpp
barulicm: eh 🤷 it's probably fine to just use the distance from the rules

barulicm:
> I'm okay with this for now and think the logic makes sense, but I think we need to make actual tests for this before comp to verify.

Agreed on having tests. Pretty much everything in Kenobi would benefit from tests.

PR #205, opened 2024-01-25
Adds command timeout to sim radio bridge.

barulicm:
grsim doesn't have a builtin command timeout, so when Kenobi stops sending command velocities for a robot, it just keeps driving at whatever the last commanded velocity was. This is why our sim robots can wander off when they're supposed to be standing still.

This PR adds a command timeout to the sim radio bridge similar to the one in our physical radio bridge. If commands stop coming in for a robot, the bridge will now send zero'd out commands to grsim to stop the robot.


fourpenny (approved):
The robots will stop moving in sim now and this passes our tests. Everything looks fine to me code/style-wise, so I think this is okay to merge.

fourpenny:
There is a lot of other wonky behavior I'm seeing in sim (wall play was very interesting to watch), but I assume that's because of path planning and some other unrelated issues we need to fix.

PR #206, opened 2024-01-27
Adds geometry testing utils.

barulicm:
While working on path planning tests, I realized we're probably going to want to regularly test if two points are near each other (mostly to handle floating point equality comparisons).

This PR adds two gtest matchers that check if points are within some threshold distance of each other.

To check if one point is near a target:

```C++
EXPECT_THAT(p1, PointIsNear(target));
```

To check if a sequence of points (ie. a path) matches another:

```C++
EXPECT_THAT(path, testing::Pointwise(PointsAreNear(), expected_path));
```

### DIFF
--- a/ateam_geometry/CMakeLists.txt
+++ b/ateam_geometry/CMakeLists.txt
@@ -39,6 +39,22 @@ install(TARGETS ${PROJECT_NAME}
         INCLUDES DESTINATION include)
 
 if(BUILD_TESTING)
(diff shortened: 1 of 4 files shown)


fourpenny (approved):
Properly builds and runs, plus is a nice utility to use in the future. Should we document info on how to use `testing::Pointwise` with the `PointsAreNear()` matcher somewhere? Or are we assuming the examples in the tests are suitable enough for someone else who might need to make tests to read/copy from for future testing?

PR #210, opened 2024-01-28
Disables ateam_ai package from building.

barulicm:
We're not using ateam_ai right now. Disabling it should save us a bit of time on our CI builds and makes it so we don't have to worry about making breaking changes to other packages used by ateam_ai.


fourpenny (approved):
good

PR #211, opened 2024-01-29
Set individual PID gains

fourpenny:
Convenient functions to set all or individual gain values for our robots' PID controllers.

Added so we can set these values in individual plays/skills when needed.


barulicm (approved):

chachmu (approved):
LGTM

PR #220, opened 2024-02-12
Hungarian algorithm max cost assignment

fourpenny:
New implementation of the Hungarian assignment algorithm ("km_assignment") utilizing a [bipartite matching implementation](https://cse.hkust.edu.hk/~golin/COMP572/Notes/Matching.pdf) (rather matrix formulation) + tests. I stole most of the code from [dlib](https://github.com/davisking/dlib/blob/master/dlib/optimization/max_cost_assignment.h) with some minor changes to use float values for costs, squarize cost matrices, and ensure non-existent goals are not assigned.

We can successfully use this implementation to select the max value assignment in test cases with asymmetric cost matrices, costs of 0, and negative costs.

This doesn't actually replace the existing greedy or Hungarian assignment implementation in our actual code to assign robots. (But I can add that to this PR if we want.)


CollinAvidano:
Are we sure that using float costs and an augmenting path this can't get into the same situation ford fulkerson can encounter with never converging.

ateam_common/include/ateam_common/km_assignment.hpp
```diff
@@ -0,0 +1,44 @@
...
+inline void compute_slack(
+  const int x,
+  std::vector<double> & slack,
+  std::vector<double> & slackx,
+  const Eigen::MatrixXd & cost,
+  const Eigen::VectorXd & lx,
+  const Eigen::VectorXd & ly
+);
+
+std::vector<int> max_cost_assignment(
```
barulicm, line 40: Can we add some documentation comments on these functions at least to give info on the structure and meaning of the inputs and outputs?

ateam_common/include/ateam_common/km_assignment.hpp
```diff
@@ -0,0 +1,44 @@
...
+
+#include <Eigen/Dense>
+#include <vector>
+
+namespace ateam_common::km_assignment
+{
+const Eigen::MatrixXd make_square_cost_matrix(
+  const Eigen::MatrixXd & cost_matrix);
+
+inline void compute_slack(
```
barulicm, line 31: `inline` likely does nothing here.

```suggestion
void compute_slack(
```

ateam_common/src/km_assignment.cpp
```diff
@@ -0,0 +1,262 @@
...
+  if (matrix.rows() == matrix.cols()) {
+    return matrix;
+  }
+
+  std::size_t new_dim = std::max(matrix.rows(), matrix.cols());
+
+  //  The value of any non-existant cost is 0 (since we are
+  //  trying to find a MAX cost matching)
+  Eigen::MatrixXd new_mat =
+    Eigen::MatrixXd::Constant(new_dim, new_dim, std::numeric_limits<double>::min());
```
barulicm, line 43: Should we change this to use `std::numeric_limits<double>::lowest()` so we get the most negative cost possible?

Otherwise, these "empty" cost cells are better assignments costwise than negative cost assignments in the original matrix.

ateam_common/src/km_assignment.cpp
```diff
@@ -0,0 +1,262 @@
...
+{
+  auto cost = cost_matrix;
+  if (cost_matrix.cols() != cost_matrix.rows()) {
+    cost = make_square_cost_matrix(cost_matrix);
+  }
+  /*
+      The dlib implementation of this algorithm is based on info available below...
+      although the links were broken when I tried them.
+          http://www.math.uwo.ca/~mdawes/courses/344/kuhn-munkres.pdf
+          http://www.topcoder.com/tc?module=Static&d1=tutorials&d2=hungarianAlgorithm
```
barulicm, line 79: We can probably omit these since they are indeed long dead links. It's unfortunate because that uwo.ca link seems to be referenced in a lot of places online.

ateam_common/src/km_assignment.cpp
```diff
@@ -0,0 +1,262 @@
...
+  std::vector<char> S, T;
+  std::vector<double> slack;
+  std::vector<double> slackx;
+  std::vector<int> aug_path;
+
+  size_t cost_size = static_cast<size_t>(cost.cols());
+
+  // Initially, nothing is matched.
+  xy.assign(cost_size, -1);
+  yx.assign(cost_size, -1);
```
barulicm, line 97: This step could just be done in the initialization for these variables.

```
std::vector<int> xy(cost_size, -1);
std::vector<int> yx(cost_size, -1);
```

ateam_common/include/ateam_common/km_assignment.hpp
barulicm, line 40: Have we tested what happens if we accidentally give this a NaN cost entry?

fourpenny:
> Are we sure that using float costs and an augmenting path this can't get into the same situation ford fulkerson can encounter with never converging.

Not sure- I'd like to do some more robust testing (and probably need to do some reading on any particular edge cases where this might not converge) if you have ideas to start with. I know you're somewhat familiar with what was causing this in the vision filter so it would probably be ideal to start with that.

ateam_common/test/test_km_assignment.cpp
```diff
@@ -0,0 +1,144 @@
...
+    {1, 0.005}
+  };
+  std::vector<int> out2x2 = km_assignment::max_cost_assignment(
+    cost2x2, false);
+  EXPECT_EQ(out2x2.size(), 2);
+  EXPECT_EQ(out2x2.at(0), 0);
+  EXPECT_EQ(out2x2.at(1), 1);
+}
+
+TEST(KmAssignment, handle_nan) {
```
fourpenny, line 124: @barulicm made this test case and addressed some problems that came up

fourpenny:
> Something with changes to numeric limits (to ::lowest and when changing to negative ::infinity) makes some tests fail, need to look at the initialization of slack and `compute_slack` more generally to see why...

Refactored to scale costs to all be positive and non-valid costs to be zero.

If we want the min cost assignment, we change the signs of the costs, then rescale, then set invalid agents/tasks to have 0 cost.

ateam_common/include/ateam_common/km_assignment.hpp
fourpenny: Should be more comments now!

fourpenny:
I'm working on replacing the existing assignment code in the multi-hypothesis Kalman filter with this to see if I can get it to work, but we can merge this first if that's better

ateam_common/src/km_assignment.cpp
```diff
@@ -0,0 +1,331 @@
...
+    https://cp-algorithms.com/graph/hungarian-algorithm.html#implementation-of-the-hungarian-algorithm
+  */
+
+  // Make sure our matrix fits our algorithms requirements
+  // (it is square and non-negative).
+  auto cost = cost_matrix;
+  if (cost_matrix.cols() != cost_matrix.rows()) {
+    cost = make_square_cost_matrix(cost_matrix);
+  }
+  cost = scale_matrix(cost, max_cost);
```
barulicm, line 146: Do we need to flip the order of these steps?
If a non-square cost matrix comes in with both negative and positive numbers, the zeros added by `make_square_cost_matrix` will get shifted up to positive costs by `scale_matrix`. If the intent is that the zeros added by the squaring function are to indicate "don't pick me unless you have to," then we probably want to add those after the scaling shifts the costs to be non-negative.

barulicm (commented):
I'm on board with merging this before doing all the work to get it integrated into the vision filter. That could let us work on integration in Kenobi in parallel.

I think this is pretty close to ready as is and is looking good.

ateam_common/src/km_assignment.cpp
fourpenny: Yes, we do need to flip these! I addressed a similar issue with the order of implementation for dealing with NaN values but forgot to check it for the general case, and my test cases were simple enough that this didn't cause any issues.

ateam_common/include/ateam_common/km_assignment.hpp
```diff
@@ -0,0 +1,114 @@
...
+{
+/*
+  Ensure the input cost matrix is non-negative. If not "is_max_cost",
+  multiply the matrix by -1 so the maximum is a minimum and vice versa
+  (since this implementation of the Hungarian algorithm finds the
+  MAX value matching).
+
+  Returns a new version of the input matrix scaled from 0 to inf.
+*/
+const Eigen::MatrixXd scale_matrix(
```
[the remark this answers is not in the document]
fourpenny: Thanks, fixed!

barulicm (approved):

PR #224, opened 2024-02-22
Adds namespace for vision filter.

barulicm:
Intellisense in vs code was being annoying when working on kenobi because it couldn't tell the difference between kenobi's types (ie. World, Robot, etc.) and vision filter's types since they had the same name and the vision filter code didn't define a namespace. This PR puts everything in the vision filter behind a namespace. This makes no functional changes to the filter itself.


fourpenny (commented):
Do you want to use the ateam_vision_filter name space for the `models.hpp` file in `types/`?

fourpenny (approved):

barulicm:
> Do you want to use the ateam_vision_filter name space for the `models.hpp` file in `types/`?

Not sure how I missed that. Fixed.

PR #225, opened 2024-03-08
GameControllerListener Quals Changes

barulicm:
This PR:

- Adds tracking of previous game command into the GC listener
- Exposes full latest ref message via getter
- Refactors parsing game command into a function for readability

(These are changes we made in prep for quals)


ateam_common/src/game_controller_listener.cpp
```diff
@@ -106,91 +108,81 @@ void GameControllerListener::RefereeMessageCallback(
 
   game_stage_ = static_cast<GameStage>(msg->stage);
 
-  uint8_t command = msg->command;
+  game_command_ = ConvertGameCommand(msg->command);
+  if (game_command_ != cached_game_command_) {
+    prev_game_command_ = cached_game_command_;
+    cached_game_command_ = game_command_;
+  }
```
fourpenny, line 115: What's the reason we cache the game command?

barulicm: Hmmm we don't actually need that to be a distinct variable I think. I can massage this code a bit to get rid of that while keeping the same behavior.

PR #226, opened 2024-03-08
Adds double touch evaluator

barulicm:
Adds the double touch evaluator from the quals branch.

This PR doesn't add the usage side of this in the kenobi gameplay code. That will come in a future PR.

### DIFF
--- a/ateam_kenobi/src/double_touch_eval.hpp
+++ b/ateam_kenobi/src/double_touch_eval.hpp
@@ -1,0 +1,122 @@
+// Copyright 2021 A Team
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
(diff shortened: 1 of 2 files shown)


ateam_kenobi/src/double_touch_eval.hpp
```diff
@@ -0,0 +1,122 @@
...
+
+    if (prev_touching_id_ && !maybe_toucher) {
+      // A robot has stopped touching the ball
+      if (!forbidden_id_) {
+        // This was the first robot to touch the ball, it can't touch it again
+        forbidden_id_ = prev_touching_id_;
+      }
+    }
+
+    world.double_touch_forbidden_id_ = forbidden_id_;
```
fourpenny, line 85: I assume we use this downstream after it is set? Where is it ultimately used?

ateam_kenobi/src/double_touch_eval.hpp
```diff
@@ -0,0 +1,122 @@
...
+  {
+    auto found_iter = std::ranges::find_if(
+      world.our_robots, [this, &world](const std::optional<Robot> & maybe_robot) {
+        if (!maybe_robot) {
+          return false;
+        }
+        return std::sqrt(
+          CGAL::squared_distance(
+            maybe_robot.value().pos,
+            world.ball.pos)) < (kRobotRadius + 0.025);
```
fourpenny, line 111: Do we want to make this configurable/not a magic number?

fourpenny (approved):
See questions above

PR #229, opened 2024-03-26
Kenobi quals changes

barulicm:
This PR brings in all of the remaining changes to Kenobi that we made while preparing for 2024 qualification submissions.

### DIFF
--- a/ateam_kenobi/src/kenobi_node.cpp
+++ b/ateam_kenobi/src/kenobi_node.cpp
@@ -110,8 +110,6 @@ private:
   PlaySelector play_selector_;
   InPlayEval in_play_eval_;
   DoubleTouchEval double_touch_eval_;
-  ateam_common::GameCommand cached_game_command_;
-  ateam_common::GameCommand prev_game_command_;
   rclcpp::Publisher<ateam_msgs::msg::OverlayArray>::SharedPtr overlay_publisher_;
   rclcpp::Publisher<ateam_msgs::msg::PlayInfo>::SharedPtr play_info_publisher_;
   rclcpp::Subscription<ateam_msgs::msg::BallState>::SharedPtr ball_subscription_;
--- a/ateam_kenobi/src/kenobi_node.cpp
+++ b/ateam_kenobi/src/kenobi_node.cpp
@@ -41,6 +41,7 @@
 #include "types/message_conversions.hpp"
 #include "play_selector.hpp"
 #include "in_play_eval.hpp"
+#include "double_touch_eval.hpp"
 #include "motion/world_to_body_vel.hpp"
 
 namespace ateam_kenobi
@@ -108,6 +109,9 @@ private:
   World world_;
   PlaySelector play_selector_;
   InPlayEval in_play_eval_;
+  DoubleTouchEval double_touch_eval_;
+  ateam_common::GameCommand cached_game_command_;
+  ateam_common::GameCommand prev_game_command_;
   rclcpp::Publisher<ateam_msgs::msg::OverlayArray>::SharedPtr overlay_publisher_;
   rclcpp::Publisher<ateam_msgs::msg::PlayInfo>::SharedPtr play_info_publisher_;
   rclcpp::Subscription<ateam_msgs::msg::BallState>::SharedPtr ball_subscription_;
@@ -218,11 +222,17 @@ private:
   {
     world_.current_time = std::chrono::steady_clock::now();
     world_.referee_info.running_command = game_controller_listener_.GetGameCommand();
+    world_.referee_info.command_time = std::chrono::system_clock::time_point(std::chrono::nanoseconds(rclcpp::Time(game_controller_listener_.GetLatestRefereeMessage().command_timestamp).nanoseconds()));
+    world_.referee_info.prev_command = game_controller_listener_.GetPreviousGameCommand();
     world_.referee_info.current_game_stage = game_controller_listener_.GetGameStage();
     if (game_controller_listener_.GetOurGoalieID().has_value()) {
       world_.referee_info.our_goalie_id = game_controller_listener_.GetOurGoalieID().value();
     }
+    if (game_controller_listener_.GetTheirGoalieID().has_value()) {
+      world_.referee_info.their_goalie_id = game_controller_listener_.GetTheirGoalieID().value();
+    }
     in_play_eval_.Update(world_);
+    double_touch_eval_.update(world_);
     if (game_controller_listener_.GetTeamColor() == ateam_common::TeamColor::Unknown) {
       auto & clk = *this->get_clock();
       RCLCPP_WARN_THROTTLE(
--- a/ateam_kenobi/src/kenobi_node.cpp
+++ b/ateam_kenobi/src/kenobi_node.cpp
@@ -222,7 +222,11 @@ private:
   {
     world_.current_time = std::chrono::steady_clock::now();
     world_.referee_info.running_command = game_controller_listener_.GetGameCommand();
-    world_.referee_info.command_time = std::chrono::system_clock::time_point(std::chrono::nanoseconds(rclcpp::Time(game_controller_listener_.GetLatestRefereeMessage().command_timestamp).nanoseconds()));
+    world_.referee_info.command_time =
+      std::chrono::system_clock::time_point(
+      std::chrono::nanoseconds(
+        rclcpp::Time(
+          game_controller_listener_.GetLatestRefereeMessage().command_timestamp).nanoseconds()));
     world_.referee_info.prev_command = game_controller_listener_.GetPreviousGameCommand();
     world_.referee_info.current_game_stage = game_controller_listener_.GetGameStage();
     if (game_controller_listener_.GetOurGoalieID().has_value()) {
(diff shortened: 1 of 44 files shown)


ateam_kenobi/src/plays/basic_122.cpp
```diff
@@ -51,63 +53,133 @@ std::array<std::optional<ateam_msgs::msg::RobotMotionCommand>, 16> Basic122::run
...
+
+  play_info_["Possession"] = they_have_possession ? "theirs" : "ours";
+
+  if (they_have_possession) {
+    const auto ball_to_bot_vec = striker_bot.pos - world.ball.pos;
+    const auto vel = ateam_geometry::normalize(ball_to_bot_vec) * 0.25;
+    motion_command.twist.linear.x = vel.x();
+    motion_command.twist.linear.y = vel.y();
     return;
   }
```
fourpenny, line 110: I think this logic is a good potential target for next easy gameplay behaviors to improve

ateam_kenobi/src/play_helpers/robot_assignment.hpp
```diff
@@ -0,0 +1,39 @@
...
+#include <optional>
+#include "types/robot.hpp"
+
+namespace ateam_kenobi::play_helpers
+{
+
+std::vector<std::optional<Robot>> assignRobots(
+  const std::vector<Robot> & available_robots,
+  const std::vector<ateam_geometry::Point> & positions,
+  const std::vector<std::vector<int>> & disallowed_robot_ids = {});
```
fourpenny, line 35:
```suggestion
/*
  Greedily assign robots to available positions
*/
std::vector<std::optional<Robot>> assignRobots(
  const std::vector<Robot> & available_robots,
  const std::vector<ateam_geometry::Point> & positions,
  const std::vector<std::vector<int>> & disallowed_robot_ids = {});
```

ateam_kenobi/src/skills/blockers.cpp
```diff
@@ -101,6 +105,6 @@ std::vector<Robot> Blockers::getRankedBlockableRobots(const World & world)
 ateam_geometry::Point Blockers::getBlockingPosition(const World & world, const Robot & blockee)
 {
   return blockee.pos +
-         ((kRobotDiameter + 0.1) * ateam_geometry::normalize(world.ball.pos - blockee.pos));
+         ((kRobotDiameter * 2.5) * ateam_geometry::normalize(world.ball.pos - blockee.pos));
```
fourpenny, line 108: Magic number??

ateam_kenobi/src/skills/goalie.cpp
```diff
@@ -52,18 +59,207 @@ void Goalie::runFrame(
...
+  easy_move_to_.setTargetPosition(
+    ateam_geometry::NearestPointOnSegment(
+      goalie_line,
+      shot_point_on_extended_goalie_line));
+  easy_move_to_.face_point(world.ball.pos);
+
+  return easy_move_to_.runFrame(goalie, world);
+}
+
+ateam_msgs::msg::RobotMotionCommand Goalie::runBlockBall(const World & world, const Robot & goalie)
```
fourpenny, line 202: We can probably refactor `runBlockBall` and `runBlockShot` a bit in the future since they share similar logic

ateam_kenobi/test/path_planner_test.cpp
```diff
@@ -177,8 +199,9 @@ TEST_F(GetPathTest, FailIfNotIgnoringStartObstacles) {
   runTest();
 }
 
-TEST_F(GetPathTest, LoopsShouldBeRemoved)
+TEST_F(GetPathTest, DISABLED_LoopsShouldBeRemoved)
 {
+  // We broke this test at some point during quals prep.
```
fourpenny, line 204: Hm, should we add a TODO or github issue for this? Or not urgent for now?

ateam_kenobi/src/plays/halt_play.cpp
```diff
@@ -33,7 +33,7 @@ void HaltPlay::reset()
 }
 
 std::array<std::optional<ateam_msgs::msg::RobotMotionCommand>, 16> HaltPlay::runFrame(
-  const World & world)
+  const World &)
```
fourpenny, line 36: did you mean to remove the reference to `world` here?

ateam_kenobi/src/skills/blockers.cpp
barulicm: Definitely a magic number we picked from testing and it just determines the distance the blocker robot should keep from the opponent.

ateam_kenobi/src/plays/halt_play.cpp
barulicm: I removed the parameter name because we don't use that parameter in this function and this is the normal way to suppress unused parameter warnings from the compiler.

ateam_kenobi/src/plays/basic_122.cpp
barulicm: Definitely agree

ateam_kenobi/src/kenobi_node.cpp
```diff
@@ -108,6 +109,9 @@ class KenobiNode : public rclcpp::Node
   World world_;
   PlaySelector play_selector_;
   InPlayEval in_play_eval_;
+  DoubleTouchEval double_touch_eval_;
+  ateam_common::GameCommand cached_game_command_;
```
fourpenny, line 113: It doesn't appear that we use this `cached_game_command_` anywhere that I can find

ateam_kenobi/src/kenobi_node.cpp
```diff
@@ -218,11 +222,21 @@ class KenobiNode : public rclcpp::Node
...
+    world_.referee_info.prev_command = game_controller_listener_.GetPreviousGameCommand();
     world_.referee_info.current_game_stage = game_controller_listener_.GetGameStage();
     if (game_controller_listener_.GetOurGoalieID().has_value()) {
       world_.referee_info.our_goalie_id = game_controller_listener_.GetOurGoalieID().value();
     }
+    if (game_controller_listener_.GetTheirGoalieID().has_value()) {
+      world_.referee_info.their_goalie_id = game_controller_listener_.GetTheirGoalieID().value();
+    }
     in_play_eval_.Update(world_);
+    double_touch_eval_.update(world_);
```
fourpenny, line 239: We should probably pick some more specific style guidelines beyond our automated tests/linter at some point. It's probably unideal to have some  classes with an `update` and others with an `Update` method...

ateam_kenobi/src/kenobi_node.cpp
barulicm: huh, looks like cached and prev are not used. Maybe these were leftover from some temporary work.

ateam_kenobi/src/play_helpers/robot_assignment.hpp
barulicm: The greedy part is more of an implementation detail that will go away when I can rewrite this to use your new assignment algorithm.

ateam_kenobi/src/kenobi_node.cpp
barulicm: removed

ateam_kenobi/src/kenobi_node.cpp
barulicm: I think that only does formatting, which we already have. I don't think clang-format can enforce symbol names.

ateam_kenobi/src/path_planning/path_planner.cpp
```diff
@@ -352,4 +372,50 @@ void PathPlanner::trimPathAfterCollision(
...
+  const auto direction_vector = start - goal;
+  const auto segment_length = std::sqrt(direction_vector.squared_length());
+  const auto step_vector = direction_vector * (options.collision_check_resolution / segment_length);
+  const int step_count = segment_length / options.collision_check_resolution;
+  for (auto step = 0; step < step_count; ++step) {
+    const auto state = goal + (step * step_vector);
+    if (isStateValid(state, world, obstacles, options)) {
+      return state;
+    }
+  }
```
fourpenny, line 389: Should the direction_vector here be `goal - start` if we start at the goal in the loop starting at 384 and go backwards towards the start?

barulicm: `start - goal` gives the vector from `goal` to `start`.
One way to sanity check this is to look at the end iteration for the loop. `step * step_vector` will equal `start - goal`. Doing the substitution on the initializer for `state`, you get `goal + start - goal`, which reduces down to just `start`, which is where we wanted to end up.
If we flipped it to `goal - start`, that end case would result in `goal + goal - start` or `(2 * goal) - start`.

fourpenny (approved):
Let's get these quals changes in :)

ateam_kenobi/src/play_helpers/robot_assignment.hpp
fourpenny: Sounds good

ateam_kenobi/src/skills/blockers.cpp
fourpenny: Also probably another good piece of logic to look at updating/understanding in that case

ateam_kenobi/src/plays/halt_play.cpp
fourpenny: Ah sorry, forgot this inherits from BasePlay!

ateam_kenobi/src/kenobi_node.cpp
fourpenny: Darn

PR #230, opened 2024-04-03
Imports UI changes from quals branch.

barulicm:
These are the changes we made to the UI package on the quals branch.

### DIFF
--- a/ateam_ui/src/src/components/FieldComponent.vue
+++ b/ateam_ui/src/src/components/FieldComponent.vue
@@ -17,8 +17,8 @@ export default {
     mounted() {
         // size is based on 140 pixels per meter times the standard max field size including boundaries (13.4 x 10.4m)
         // TODO: Make this properly resize itself
-        const width = 1876;
-        const height = 1456;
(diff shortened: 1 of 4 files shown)


ateam_ui/src/src/components/FieldComponent.vue
```diff
@@ -17,8 +17,8 @@ export default {
     mounted() {
         // size is based on 140 pixels per meter times the standard max field size including boundaries (13.4 x 10.4m)
         // TODO: Make this properly resize itself
-        const width = 1876;
-        const height = 1456;
+        const width = 1876*0.66;
```
chachmu, line 20: I think these lines were just changed to make it look better on the coach computer's high resolution screen. We can keep this or remove it, either way is fine

chachmu (approved):
LGTM

ateam_ui/src/src/components/FieldComponent.vue
barulicm: Removed.

PR #231, opened 2024-04-03
Removes Kenobi files deleted in quals branch.

barulicm:
These deletes were missed when I was importing changes from the quals branch in the previous Kenobi PR.


fourpenny (approved):
Since things still built, I assume removing these files was fine

PR #232, opened 2024-04-10
Quick assignment fix and addition of unassignable positions

fourpenny:
Haven't added tests for this yet, but wanted to make sure it could get merged so we can use a common implementation for both Matt and I's work that will use the new km_assignment functions.


ateam_common/src/km_assignment.cpp
```diff
@@ -88,6 +87,25 @@ Eigen::MatrixXd replace_nan_costs_with_value(
   return new_matrix;
 }
 
+Eigen::MatrixXd replace_forbidden_costs_with_zeros(
```
barulicm, line 90: How is this intended to work?
As I read this, you would be zero-ing out the costs of all tasks for a given robot or vice versa.
But we want to be able to disallow specific robots for specific tasks while allowing those robots to be accurately costed for other tasks.
Also, is there code somewhere that checks for zero cost assignments, or would this still assign disallowed robots if no other bots were available?

ateam_common/src/km_assignment.cpp
```diff
@@ -115,7 +133,9 @@ value.
 
 std::vector<int> max_cost_assignment(
   const Eigen::MatrixXd & cost_matrix,
-  bool max_cost
+  bool max_cost, 
+  std::vector<int> forbidden_x, 
+  std::vector<int> forbidden_y
```
barulicm, line 138: It's a little weird that the max and min versions of this function have different signatures. Did an issue come up with the enum idea we had the other night?

fourpenny:
Sorry @barulicm, I honestly was really brain dead when making these changes after work and totally misinterpreted the notes I wrote down for myself about fixes you suggested last week. Going to start addressing them tonight

ateam_common/src/km_assignment.cpp
fourpenny: I think this new implementation is more logical, let me know if you see any clear errors with it.

I think there does need to be more robust checking to see if a particular cost is zero and explicitly not assign those tasks... I need to add some test cases that purposefully fail due to lack of assignable tasks/robots and handle them; right now, I think the assignment function will continue endlessly if there is no actual solution.

ateam_common/src/km_assignment.cpp
fourpenny: Should be changed to use an enum now

ateam_common/src/km_assignment.cpp
barulicm: When we're doing min cost assignment and the cost matrix gets scaled, wouldn't the originally max cost robot-assignment pair be set to zero? If we go forward with "zero cost means disallowed", then that robot would become disallowed unexpectedly.
Is my understanding here correct?

barulicm:
Can we add a test for the specific case we saw fail in the kenobi test suite? (TwoRobotsOneGoal)
Specifically, the cost matrix was: 
```
[ 2.8 ]
[ 0   ]
```
And the expected min-cost assignment was to select the second robot.

fourpenny:
> Can we add a test for the specific case we saw fail in the kenobi test suite? (TwoRobotsOneGoal) Specifically, the cost matrix was:
> 
> ```
> [ 2.8 ]
> [ 0   ]
> ```
> 
> And the expected min-cost assignment was to select the second robot.

This is done, along with some additional tests

ateam_common/src/km_assignment.cpp
barulicm: Actually looks like I misread what the change was.
I think the above issue might still be a problem? Though I'm also not sure what the role of the code below that sets entries in `xy` to -1 is.

barulicm (approved):

PR #233, opened 2024-04-19
Big geometry update

barulicm:
A couple big changes to the geometry library:

- Adds Arc type to represent circular arcs
- Adds Disk type as a semantic type mainly for intersections.
  Allows us to distinguish between intersections on the circle from intersections on and in the circle.
- Replaces our strategy of adding functions for certain types (ala. `variantDoIntersect`) with more general wrapper functions that forward to CGAL when able (`ateam_geometry::doIntersect` and `ateam_geometry::intersection`)
- Adds a handful of small utility functions
- Updates copyright year to 2024
- Makes all function names match [ROS C++ style guides](https://wiki.ros.org/CppStyleGuide)


ateam_kenobi/src/path_planning/path_planner.cpp
```diff
@@ -51,7 +51,7 @@ PathPlanner::Path PathPlanner::getPath(
   }
 
   if (options.avoid_ball) {
-    augmented_obstacles.push_back(ateam_geometry::makeCircle(world.ball.pos, 0.04267 / 2));
+    augmented_obstacles.push_back(ateam_geometry::makeDisk(world.ball.pos, 0.04267 / 2));
```
chachmu, line 54: While we are modifying these files we might as well change this constant over to `kBallDiameter`

chachmu (approved):
Looks great!

PR #238, opened 2024-05-12
Improve Stop play smarts

barulicm:
Our existing stop behavior of blindly driving away from the ball can lead to collisions and other bad behavior.
This new play looks for open spots around the ball and moves bots to the open spots.

This included a couple drive by changes:
- Add arcs to overlay system
- Fix a few things in the Arc type in the geometry package
- Fix bad math in window evaluation shadow code and refactor so other teams can use it too

Note: Moving robots to the _closest_ open spot depends on future improvements to our assignment system in Kenobi.


chachmu (approved):
👍

PR #239, opened 2024-05-15
KM Assignment in Kenobi

barulicm:
Rewrites the robot assignment play helper in Kenobi to use the new KM assignment implementation.


fourpenny (approved):
Looks like a drag and drop replacement for our previous assignment- builds and tests pass for me :)

PR #240, opened 2024-05-16
Play selection overhaul

barulicm:
Some big changes to our play selection system:

- Plays now can override `getScore()` to tell the play selection system when they should be used. Plays can rank themselves based on the world state so we can start adding multiple play options for each game state.
- Play selection can be overridden with the `/kenobi_node/set_override_play` service. The selected play will be picked regardless of play scores. The Halt referee command will always run the halt play.
- Plays can be enabled / disabled with the `/kenobi_node/set_play_enabled` service. Disabled plays will be skipped in the play ranking process.
- The `/kenobi_node/get_play_names` service will return the list of play names. This will help us port these new play controls into the UI.


ateam_kenobi/src/play_selector.cpp
```diff
@@ -21,84 +21,123 @@
 
 #include <iostream>
 #include "play_selector.hpp"
+#include "plays/all_plays.hpp"
 #include "ateam_common/game_controller_listener.hpp"
 
 namespace ateam_kenobi
 {
 
+PlaySelector::PlaySelector()
```
chachmu, line 30: Just out of curiosity, is there an easy way long term to get plays automatically added to this list and the CMake? It seems like we get slowed down a lot trying to use plays that aren't getting compiled.

chachmu (approved):
Looks great!

ateam_kenobi/src/play_selector.cpp
barulicm: I think so. I was giving some thought to that too while writing this. A bit of CMake + macro magic should be doable.

PR #241, opened 2024-05-20
STP Infrastructure

barulicm:
This PR adds STP (skill, tactic, play) base types that allow construction of a sort of tree of gameplay elements.

The primary goal is to, as easily as possible, enable hierarchical namespacing of things like logs, parameters, and overlays throughout our gameplay structures. I wanted STP objects to remain as members of their owning classes to make it easy to interact with them. I also wanted it to be as painless as possible to create STP "children," passing the core utilities down with extended namespaces.

Everything in this PR follows from the creation of the `stp::Base` class. This is the base type for all STP objects. It owns the underlying resources for the core utilities and provides functions for creating children.
Three additional STP types inherit directly from `stp::Base`: `stp::Play`, `stp::Tactic`, and `stp::Skill`. These allow us to add members / functionality common to each class of object. `stp::Play` is the only one that really does anything for now.
It is assumed that `stp::Play` objects will form the roots of the STP tree in our system, so they get some special treatment in some spots.

Classes that are not skills, tactics, or plays can still be STP-aware. `EasyMoveTo` and `PathPlanner` are done in this PR. They now inherit from `stp::Base` which allows them to easily use namespaced logging, overlays, etc.

Building the STP tree boils down to inheriting from an STP base type and calling `createChild` or `createIndexedChildren` to initialize your STP children. `createIndexedChildren` just calls `createChild` 16 times, appending the index to the child name.

This PR also introduces ROS-based logging to the new STP infrastructure. Any STP object can now use the rclcpp logging macros and the `getLogger()` function to log through ROS. Log messages will be labelled with the STP namespace for the object calling the log macro.

The next steps in this line of work is to add a parameter management system to the STP system.


ateam_kenobi/src/play_selector.cpp
```diff
@@ -30,27 +30,29 @@
 namespace ateam_kenobi
 {
 
-PlaySelector::PlaySelector()
+PlaySelector::PlaySelector(rclcpp::Node & node)
```
chachmu, line 33: Would it make more sense to just pass the logger in here?

chachmu (approved):
Looks awesome

ateam_kenobi/src/play_selector.cpp
barulicm: I think I'll need the node in the next wave of work to add a parameter management system.

PR #243, opened 2024-05-29
Adds Disk support to PathPlanner::drawObstacles.

barulicm:
This should quiet the noisy warnings from `PathPlanner` about Disks.


chachmu (approved):
This one took a while to review but I think it should be good 👍

PR #246, opened 2024-06-03
STP Parameters

barulicm:
Adds infrastructure to support ROS parameters in STP objects.

Parameters in this system are namespaced to their part of the STP tree, rooted with "stp_parameters".

So, for example, the only parameter added in this PR is named `stp_parameters.TestKickPlay.use_pivot_kick`, and corresponds to this tree:

- stp root
  - TestKickPlay
    - use_pivot_kick


chachmu (approved):
Looks good. I can probably find a way to get the parameters to show up in the UI too so we can tweak play settings on the fly. I guess the only concern would be that they probably get reset if kenobi crashes and restarts so we would have to make sure to have the correct defaults compiled in before matches but it would be very nice for testing.

barulicm:
We can also trigger the `ros2 param dump` command from a button to save the current state to a file.

PR #247, opened 2024-06-10
Add Param List UI element

chachmu:
When first opening the param list give it 3 seconds to fetch the params from ROS. If it still hasn't populated the values or you want to refetch the values use the refresh button at the top of the panel.

### DIFF
--- a/ateam_ui/src/src/state.ts
+++ b/ateam_ui/src/src/state.ts
@@ -7,6 +7,7 @@ import { Ball } from "@/ball"
 import { Field, FieldDimensions, FieldSidedInfo } from "@/field"
 import { AIState } from "@/AI"
 import { Play } from "@/play"
+import { Param } from "@/param"
 
 export class RenderConfig {
     angle: number = 0; // Rotation applied to the rendered field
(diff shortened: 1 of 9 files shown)


ateam_ui/src/src/state.ts
```diff
@@ -57,6 +58,55 @@ export class AppState {
...
+
+                    // /kenobi_node:stp_parameters.TestKickPlay.use_pivot_kick
+                    // Make sure param has play name and stp param name
+                    if (substrings.length < 3) {
+                        continue;
+                    }
+
+                    // Expected format is: "/kenobi_node:stp_parameters.PlayName.ParamName"
+                    const play_name = substrings[1];
+                    const stp_param_name = substrings[2];
```
barulicm, line 79: This appears to assume all parameters are at the play level. Parameters can be created anywhere in the STP tree, so in reality we might have something like this: `stp_parameters.StopPlay.EasyMoveTo.MaxVelocity`.
How hard would it be to show that full tree?

barulicm (approved):
There's been a lot of offline conversations about parameters and how they should work in the future. We'll deal with the details of what this feature should become when we've figured out how parameters should work on the Kenobi side.

PR #254, opened 2024-07-01
Vision filter uses new assignment

fourpenny:
Disabled failing old tests for now- they will need to be rewritten because in general they don't use values based on starting tracks (they instead assume no initialized values). We did test this IRL and the filter could handle 100 ball detections without breaking.


ateam_vision_filter/CMakeLists.txt
```diff
@@ -67,6 +67,7 @@ if(BUILD_TESTING)
   target_include_directories(${PROJECT_NAME}_test PRIVATE test src)
   target_link_libraries(${PROJECT_NAME}_test ${PROJECT_NAME})
   set_target_properties(${PROJECT_NAME}_test PROPERTIES CXX_STANDARD 17)
-endif()
+  target_compile_options(${PROJECT_NAME}_test PRIVATE -g)
```
barulicm, line 70: Instead of manually adding this kind of flag in CMake, we can just set the build type to `Debug` or `RelWithDebInfo` when we want debug symbols.

```suggestion
```

ateam_vision_filter/test/generators/transmission_probability_generator_test.cpp
barulicm, line 1: Is there a reason this file is all commented out instead of disabling the broken tests like the others?

chachmu (commented):
Everything other than what barulic mentioned seems good to me.

ateam_vision_filter/CMakeLists.txt
fourpenny: Sorry, I tried to use debug mode in VS Code and must have accidentally added this somehow

ateam_vision_filter/test/generators/transmission_probability_generator_test.cpp
fourpenny: No, that was an accident. Should now be fixed.

barulicm (approved):

PR #257, opened 2024-09-16
Imports comp changes for ateam_bringup.

barulicm:
* Adds use_emulated_ballsense parameter
* Adds field_manager to autonomy bringup (will break things until the field_manager is merged)
* Updates comp IPs from Einhoven
* Adds use_local_gc parameter
* Makes simulation bringup use only local GC
* Tweaks handling of expose_gc_to_net


barulicm:

PR #258, opened 2024-09-16
Adds designated position to GameControllerListener.

barulicm:
Importing changes for ateam_common from comp branch.


barulicm:

PR #259, opened 2024-09-16
Imports comp changes to ateam_msgs.

barulicm:
* Adds new fields to multiple message types
* Adds `SetIgnoredFieldSide` and `SendSimulatorControlPacket` service types

Depends on #260


barulicm:

PR #260, opened 2024-09-16
Imports comp changes to ssl_league_msgs.

barulicm:
* Removes all `<=1` array size constraints used for translating optional values. These caused major issues with various ROS support tools.
* Adds simulator control message types.


barulicm:

PR #261, opened 2024-09-16
Adds Field Manager

barulicm:
Importing changes from comp branch.

The Field Manager is responsible for parsing and publishing field info from vision messages. It also manages disabled field halves for practicing.

Depends on #259


barulicm:

PR #262, opened 2024-09-18
Imports radio bridge changes from comp branch.

barulicm:
Changes include:
* Extending timeout for firmware parameter changes
* Adds warning log for timed out commands


ateam_radio_bridge/src/radio_bridge_node.cpp
```diff
@@ -171,6 +171,7 @@ class RadioBridgeNode : public rclcpp::Node
       if ((std::chrono::steady_clock::now() - motion_command_timestamps_[id]) >
         command_timeout_threshold_)
       {
+        RCLCPP_WARN(get_logger(), "Robot %d command timed out. Sending zeros.", id);
```
guyfleeman, line 174: I recall this causing some confusion about what command specifically timed out. If I recall this means software is not sending commands to the radio bridge, not that sending the command to the robot timed out. Specifically in the bridge we might want to clarify this.

guyfleeman (changes requested):
somewhat pedantic request to make logging more clear

ateam_radio_bridge/src/radio_bridge_node.cpp
barulicm: Tweaked wording.

guyfleeman (approved):

PR #263, opened 2024-09-18
Parameterizes joystick kick speed.

barulicm:
Merging changes to joystick control package from comp branch.


ateam_joystick_control/src/joystick_control_node.cpp
```diff
@@ -73,6 +73,8 @@ class JoystickControlNode : public rclcpp::Node
     // controls the size of steps when changing dribbler speed
     declare_parameter<double>("dribbler_speed_step", 10.0);
 
+    declare_parameter<double>("kick_speed", 5.0);
```
guyfleeman, line 76: this seems like a high default, but I'll allow it

guyfleeman (approved):
changes are short and seem reasonable

ateam_joystick_control/src/joystick_control_node.cpp
barulicm: iirc, this default was chosen before the kick speed scaling was corrected in firmware.

PR #264, opened 2024-09-29
Imports ateam_goemetry changes from comp/2024

barulicm:
* Adds some angle helpers
  * Includes new dependency on ROS angles package
* Adds custom implementation of segment-segment intersection
* Adds orientation function
* Reorganizes testing utilities
* Adds `SegmentIsNear` testing matcher
* Fixes bug in `nearestPointOnSegment`


ateam_geometry/include/ateam_geometry/orientation.hpp
```diff
@@ -0,0 +1,52 @@
...
+};
+
+/**
+ * @brief Determines the orientation of the points in order A->B->C
+ * @note Based on https://www.geeksforgeeks.org/orientation-3-ordered-points/
+ */
+inline Orientation orientation(const Point & a, const Point & b, const Point & c)
+{
+  const auto slope_diff = (b.y() - a.y()) * (c.x() - b.x()) - (b.x() - a.x()) * (c.y() - b.y());
+  if (std::abs(slope_diff) < 1e-12) {
```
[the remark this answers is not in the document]
guyfleeman: And if it'll be used in multiple places it should probably not be a magic number

ateam_geometry/src/intersection.cpp
```diff
@@ -190,4 +191,58 @@ std::optional<std::variant<Point, std::pair<Point, Point>>> intersection(
...
+
+  if (o1 == Orientation::Colinear && o2 == Orientation::Colinear) {
+    // segments are colinear
+    struct PointWithSegmentIndex
+    {
+      Point point;
+      int segment_index;
+      bool operator<(const PointWithSegmentIndex & other) const
+      {
+        if (std::abs(point.x() - other.point.x()) < 1e-12) {
```
guyfleeman, line 222: same magic number as before for thresholding

ateam_geometry/test/intersection_test.cpp
```diff
@@ -166,3 +166,71 @@ TEST(IntersectionTests, ArcToSegment)
           PointIsNear(ag::Point(-0.707, -0.707)),
           PointIsNear(ag::Point(0.707, 0.707))))));
 }
+
```
guyfleeman, line 169: should this include a test with colinear lines with exactly one intersection point?

E.g.

```
  TEST(IntersectionTests, SegmentToSegmentColinearNoIntersection)
  {
    ag::Segment a{
      ag::Point{0, 0},
      ag::Point{1, 1}
    };
  
    ag::Segment b{
      ag::Point{1, 1},
      ag::Point{2, 2}
    };
  
      EXPECT_THAT(
      ag::intersection(a, b),
      Optional(VariantWith<ag::Point>(PointIsNear(ag::Point{1, 1}))));
  }
```

ateam_geometry/include/ateam_geometry/orientation.hpp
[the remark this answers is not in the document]
barulicm: I've centralized all the magic thresholds into a single header with named constants.

ateam_geometry/test/intersection_test.cpp
barulicm: Turns out it should because there was definitely a bug where this test didn't pass. It does now though.

guyfleeman (approved):
feedback addressed

PR #266, opened 2024-09-30
Adds simulation control to sim radio bridge

barulicm:
Changes from comp/2024

### DIFF
--- a/ateam_ssl_simulation_radio_bridge/src/message_conversions.cpp
+++ b/ateam_ssl_simulation_radio_bridge/src/message_conversions.cpp
@@ -19,6 +19,11 @@
 // THE SOFTWARE.
 
 #include "message_conversions.hpp"
+#include <tf2/convert.h>
+#include <tf2/utils.h>
+#include <tf2/LinearMath/Quaternion.h>
+#include <stdexcept>
+#include <tf2_geometry_msgs/tf2_geometry_msgs.hpp>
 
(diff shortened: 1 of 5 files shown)


ateam_ssl_simulation_radio_bridge/src/message_conversions.cpp
```diff
@@ -49,7 +54,7 @@ RobotControl fromMsg(const ateam_msgs::msg::RobotMotionCommand & ros_msg, int ro
     case ateam_msgs::msg::RobotMotionCommand::KICK_NOW:
     case ateam_msgs::msg::RobotMotionCommand::KICK_ON_TOUCH:
     case ateam_msgs::msg::RobotMotionCommand::KICK_ON_CAPTURE:
-      proto_robot_command->set_kick_speed(ros_msg.kick_speed);
+      proto_robot_command->set_kick_speed(ros_msg.kick_speed * 3.0);
```
guyfleeman, line 57: why do we do this multiplication?

barulicm: Trying to account for discrepancies between real robot kick speeds and sim robot kick speeds. 
All of this kick speed stuff is something we should double check, especially if/when we make new kicker hardware.

guyfleeman: seems okay for now then

guyfleeman (approved):

PR #270, opened 2024-10-09
Imports Kenobi changes from comp/2024.

barulicm:
This should be the last comp PR. I haven't read through all of it since comp. Realistically, it's more code than I personally feel like reviewing until something breaks. I'm sure we did a great job when writing it.


ateam_kenobi/src/skills/line_kick.hpp
```diff
@@ -24,58 +24,144 @@
...
+                %%%@@@@@@   @@%   @  @@@@@@@%@%@%%@@@%@%@#@    %@@@@@%:  @@@@@@@@%+----::.
+                  %%%@@@@@@%%%  @   :@%  @#  %@%   @@@@@  @@@%  @@@%@@@@@%@@@%%%==----:.
+                    +%%%@@@@@@@@@-  @@  @%  *%@  @ %@@%@% .@@@@  @@@@@@@@@%@%@===---:.
+                     .:#%%%@@@@@@@@@@%%     @%%     @@@@@  @@@@@@@@@@@@@%@%+====--:
+                      .:::*%%%@@@@@@@@@@@@@@@@%#@@@@%@@@@@@@@@@@@@@@%%%%++++==-.
+                       ..:::--#%%@@@@@@@@@@@@@@@@@@@@@@@@@@@@@%%%%%%##**++=.
+                         ..::---===*%@%%%%@@@@@@@@@@%@@%%@%%%%%%%##*+:
+                            .::---====+++**###%%%%%%%%%%%%###=
+                                  .:--==+++++++=-.
+*/
```
guyfleeman, line 137: this is a beautiful file with no flaws

guyfleeman (approved):
image

ateam_kenobi/src/skills/line_kick.hpp
```diff
@@ -24,58 +24,144 @@
...
   }
 
-  ateam_msgs::msg::RobotMotionCommand runFrame(const World & world, const Robot & robot);
+  double move_to_ball_velocity = 1.8;
+  double robot_perp_dist_to_ball_threshold = 0.008;
+  double angle_threshold = 0.1;
+  double kick_drive_velocity = 0.4;
+
+  // Allow the robot to push up against obstacles / other robots (slowly)
+  bool cowabunga = false;
```
fourpenny, line 79: Are we keeping cowabunga mode or nah

ateam_kenobi/src/skills/line_kick.hpp
fourpenny: sus

ateam_kenobi/src/play_helpers/easy_move_to.cpp
```diff
@@ -171,4 +183,72 @@ void EasyMoveTo::drawTrajectoryOverlay(
...
+  path_planning::AddRobotObstacles(world.their_robots, obstacles);
+  const auto robot_footprint = ateam_geometry::makeDisk(
+    robot.pos,
+    kRobotRadius + planner_options_.footprint_inflation);
+  const auto colliding_obstacle = path_planning::GetCollidingObstacle(robot_footprint, obstacles);
+  if (!colliding_obstacle) {
+    return std::nullopt;
+  }
+
+  const double safe_speed = 0.5;
```
fourpenny, line 204: I assume this won't change but maybe we want this "safe speed" defined in a single as a sort of constant?

ateam_kenobi/src/play_helpers/easy_move_to.cpp
```diff
@@ -171,4 +183,72 @@ void EasyMoveTo::drawTrajectoryOverlay(
...
+
+  auto msg_from_vector = [](const ateam_geometry::Vector & v) {
+      ateam_msgs::msg::RobotMotionCommand command;
+      command.twist.linear.x = v.x();
+      command.twist.linear.y = v.y();
+      return command;
+    };
+
+  if (std::holds_alternative<ateam_geometry::Circle>(*colliding_obstacle)) {
```
fourpenny, line 213: Have we hit any edge cases where we are close enough to hit multiple obstacles? In those cases what happens?

ateam_kenobi/src/play_helpers/easy_move_to.cpp
fourpenny, line 1: Are there any specific cases we aren't using EasyMoveTos for robot movements? (I didn't notice any briefly looking through all our plays, just wanted to verify my understanding of how things are working.)

ateam_kenobi/src/play_helpers/lanes.cpp
```diff
@@ -0,0 +1,149 @@
...
+#include "lanes.hpp"
+#include <ateam_geometry/do_intersect.hpp>
+
+namespace ateam_kenobi::play_helpers::lanes
+{
+
+ateam_geometry::Segment GetLaneLongitudinalMidSegment(const World & world, const Lane & lane)
+{
+  const auto half_field_length = world.field.field_length / 2.0;
+  const auto two_sixths_field_width = (2.0 / 6.0) * world.field.field_width;
```
fourpenny, line 30: is there a reason this is 2/6 and not 1/3?

ateam_kenobi/src/skills/line_kick.hpp
barulicm: I think we'll keep it for now. As we fix up motion control and come back to kick skills, we'll do some more experiments to see if cowabunga mode is useful.

ateam_kenobi/src/play_helpers/easy_move_to.cpp
barulicm: We almost certainly have hit that before. The code as is will pick the first one in the obstacles list. In practice, the order the obstacles are checked in is:
1. default obstacles (ie defender boxes)
2. our robots (lowest ID first)
3. their robots (lowet ID first.
The escape velocity will be generated from just that one obstacle, which may or may not be a useful velocity depending on the arrangement of the obstacles.
It's probably not worth fixing that behavior now, but definitely worth adding a TODO to remind us that this limitation exists.

ateam_kenobi/src/play_helpers/easy_move_to.cpp
barulicm: There are a handful of places that modify / create motion commands without using `EasyMoveTo` (ie. `LineKick::RunKickBall()`). This mostly happens in skills. I _think_ `Basic122` is the only play that does this (in `runStriker` to implement the backoff when opponent has possession).

ateam_kenobi/src/play_helpers/lanes.cpp
barulicm: :D I think it came from this function being derived from `GetLaneBounds` maybe? That function used 1/6 to mark the lane borders and I just made that 2/6 for the midsegments?
Kinda silly looking back at it now. I'll reduce it.

ateam_kenobi/src/play_helpers/easy_move_to.cpp
barulicm: It's a good idea. I'm not sure where else this would get used at the moment off the top of my head, but worth keeping an eye out.

ateam_kenobi/src/play_selector.cpp
fourpenny, line 1: We don't have a way to currently visualize/list all of our plays at a glance, do we? (Nor the states that trigger the state machine)
If there's a straightforward way to do it, maybe it would be useful to generate a diagram from this file or the imports in all plays to get a in-the-moment idea what we currently have running.

My thought is we (I?) can probably create a quick script/app using either Plotly or Go.js to do this if it sounds like a good idea.

barulicm: Depends on what you mean. We have the play list in the UI that lists all of the plays and shows their current evaluation scores.

Visualizing the play decision process is more complicated because it's all done through arbitrary scoring functions in the plays themselves. There is no longer a top level state machine choosing the plays.

fourpenny: Ah, gotcha. I thought that was still the case, so this doesn't seem as practical based on the current way we're evaluating.

ateam_kenobi/src/double_touch_eval.hpp
fourpenny, line 1: I've noticed that in general for the objects updated per frame we're not using a consistent style for function naming, even for the same thing, i.e. `update()` vs `Update()`. Maybe we can set the linter up to catch some of these?

fourpenny: I saw this in between this file and in_play_eval, for example

ateam_kenobi/src/double_touch_eval.hpp
```diff
@@ -101,13 +104,13 @@ class DoubleTouchEval
   std::optional<Robot> GetRobotTouchingBall(const World & world)
   {
     auto found_iter = std::ranges::find_if(
-      world.our_robots, [this, &world](const std::optional<Robot> & maybe_robot) {
-        if (!maybe_robot) {
+      world.our_robots, [this, &world](const Robot & robot) {
+        if (!robot.visible) {
```
fourpenny, line 108: I like the use of the visible robots helpers instead of std::optional for these checks :)

ateam_kenobi/src/double_touch_eval.hpp
barulicm: Yeah, we haven't been super consistent on following our style guide. I've tried to be more consistent on the newer stuff, but that sometimes makes it not match the old stuff until I get around to touching old stuff for other reasons.
Unfortunately, I do not know of a linter that can actually check stuff like this.

PR #273, opened 2024-11-12
ui-debug-argument-expose

CollinAvidano:
Just moves the argument for ui_debug through bringup simulations launch file


barulicm:
I don't think this is needed. All launch arguments are added to the global scope by default, so you can already just pass "ui_debug" on the command line and it will get it down to the UI launch file as the code was. This is why that flag was prefixed with "ui" to begin with.

PR #278, opened 2024-11-18
Refactors robot feedback topic names.

### LINKED ISSUE #277: Inconsistent robot topic mappings

barulicm:
Fixes #277

Refactors our feedback topic names for better usability in tools like plotjuggler. This puts status and motion feedback topics under the common "/robot_feedback" top level namespace.

This PR also makes the joystick only launch setup use the same topic names as the full stack bringup.


chachmu (approved):
:+1:

PR #280, opened 2024-11-21
Automatic network management

barulicm:
This PR adds automatic discovery / handling of IP addresses that we currently have to configure for each field / environment we run in. With these changes, manually specifying addresses / interfaces is longer required in normal circumstances.

This includes these features:
* `MulticastListener` gains the ability to listen on all interfaces at once if the interface address argument is left empty.
* IP address launch arguments are now blank to default to auto-discovery. They are kept so they can be used as command line overrides when needed.
* The team client reconnect service can now optionally change the game controller IP address.
* The GC multicast bridge node will now call the team client reconnect service when multicast packets are received and the team client is not connected.

### DIFF
--- a/ateam_bringup/launch/autonomy.launch.xml
+++ b/ateam_bringup/launch/autonomy.launch.xml
@@ -1,7 +1,7 @@
 <launch>
   <arg name="ssl_vision_ip" default="224.5.23.2"/>
   <arg name="ssl_vision_port" default="10006"/>
-  <arg name="ssl_vision_interface_address" default="10.193.15.132"/>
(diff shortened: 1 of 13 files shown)


guyfleeman (approved):

chachmu (approved):
:tada: Do we want any buttons in the UI to call the reconnect service or anything?

barulicm:
> 🎉 Do we want any buttons in the UI to call the reconnect service or anything?

Yes. On my mental TODO list has been a widget for the UI that shows the GC connection status, referee message rate, and has a reconnect button.

PR #283, opened 2025-01-06
Adds temporary warning suppression to motion controller.

barulicm:
The control_toolbox package fixed this warning [here](https://github.com/ros-controls/control_toolbox/commit/74510fd2d9d4bbf074c750d7172ea602b95888c8) but that has not been released yet. The next ros2 sync will probably be in about a month.

This PR suppresses the warning at the include site for now to unblock our builds while still treating other warnings as errors.


chachmu (approved):
LGTM 👍

PR #305, opened 2025-02-19
Adds spatial prototype

barulicm:
This adds the OpenCV-based prototype of Kenobi's spatial subsystem as we used in our quals branch.
The play that actually uses this subsystem will be imported in a future PR.

### DIFF
--- a/ateam_kenobi/src/spatial/layers/in_field.hpp
+++ b/ateam_kenobi/src/spatial/layers/in_field.hpp
@@ -18,8 +18,8 @@
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 // THE SOFTWARE.
 
-#ifndef LAYERS__IN_FIELD_HPP_
-#define LAYERS__IN_FIELD_HPP_
+#ifndef SPATIAL__LAYERS__IN_FIELD_HPP_
+#define SPATIAL__LAYERS__IN_FIELD_HPP_
 
 #include "spatial/spatial_layer_factory.hpp"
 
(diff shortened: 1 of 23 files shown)


ateam_kenobi/src/spatial/layers/in_field.hpp
```diff
@@ -0,0 +1,62 @@
...
+
+class InField : public SpatialLayerFactory {
+public:
+  InField()
+  : SpatialLayerFactory("InField") {}
+
+  void FillLayer(cv::Mat & layer, const World & world) override
+  {
+    SetupLayer(layer, CV_8UC1);
+    if(layer.size() == prev_size) {
```
chachmu, line 37: If for some reason the field boundary width changed but the total field size didn't this wouldn't trigger the layer to update.

ateam_kenobi/src/spatial/layers/their_defense_area_keepout.hpp
```diff
@@ -0,0 +1,71 @@
...
+class TheirDefenseAreaKeepout : public SpatialLayerFactory {
+public:
+  TheirDefenseAreaKeepout()
+  : SpatialLayerFactory("TheirDefenseAreaKeepout") {}
+
+  void FillLayer(cv::Mat & layer, const World & world) override
+  {
+    SetupLayer(layer, CV_8UC1);
+
+    if(world.field.field_length == prev_field.field_length ||
```
chachmu, line 38: Shouldn't this be checking if all of the field dimensions are still the same (as in using and instead of or)?

chachmu (approved):

ateam_kenobi/src/spatial/layers/their_defense_area_keepout.hpp
barulicm: ssshhhhhhh.....
but also, yeah. Good thing to keep in mind for the real version of this system :)

PR #311, opened 2025-02-19
Quals Play Changes

barulicm:
Imports changes to plays from the quals branch. This includes new plays, play tuning, and reorganizing the plays folder.

Will retarget to main when #310 is merged.


ateam_kenobi/src/plays/test_plays/controls_test_play.cpp
chachmu, line 1: We maybe should have used `git mv` for these so it would recognize we just moved the files

chachmu (approved):

ateam_kenobi/src/plays/test_plays/controls_test_play.cpp
barulicm: Ah I think I just screwed up pulling in the changes. So now we have two copies of the plays we moved. I'm going to remove the old ones and see if it figures it out then.

barulicm: Alright, that seems to have worked now that the files are actually being moved and not just duplicated.

barulicm:
@chachmu Do you want to re-review now that the changes to the moved plays are more obvious, or do we think this is good to go?

chachmu:
I mostly just wanted to make sure we preserved the history, I already reviewed the files themselves so I think we should be good to merge

PR #313, opened 2025-02-22
Adds playbook caching.

barulicm:
Adds the ability to load and save playbook state to disk, including autosave to a cache directory on shutdown.
For now, this just enables or disables plays by name. Not sure if we'll end up with other settings for plays in this kind of file in the future.

The autosave is saved to `~/.ateam/software/kenobi/playbook/autosave.json`.

An override playbook file can be chosen at launch time with the `kenobi_playbook` launch argument.

### DIFF
--- a/ateam_kenobi/CMakeLists.txt
+++ b/ateam_kenobi/CMakeLists.txt
@@ -17,6 +17,7 @@ find_package(control_toolbox REQUIRED)
 find_package(ompl REQUIRED)
 find_package(angles REQUIRED)
 find_package(OpenCV REQUIRED)
+find_package(nlohmann_json REQUIRED)
 
 add_library(kenobi_node_component SHARED
   src/defense_area_enforcement.cpp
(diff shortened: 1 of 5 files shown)


ateam_kenobi/CMakeLists.txt
```diff
@@ -17,6 +17,7 @@ find_package(control_toolbox REQUIRED)
 find_package(ompl REQUIRED)
 find_package(angles REQUIRED)
 find_package(OpenCV REQUIRED)
+find_package(nlohmann_json REQUIRED)
```
chachmu, line 20: huh... how did the play info stuff ever work if we weren't building with the nlohmann package...

chachmu (approved):

ateam_kenobi/CMakeLists.txt
barulicm: nlohmann-json is header only, so technically this line isn't strictly required as long as you luck out and some other dependency adds the needed root include path.

PR #316, opened 2025-02-22
Fix neutralino for Ubuntu 24.04

chachmu:
Turns out neutralino changed the name of some of the configuration options in the json file which caused some important configuration to just not get set

### DIFF
--- a/ateam_ui/src/neutralino.config.json
+++ b/ateam_ui/src/neutralino.config.json
@@ -1,10 +1,14 @@
 {
+  "$schema": "https://raw.githubusercontent.com/neutralinojs/neutralinojs/main/schemas/neutralino.config.schema.json",
   "applicationId": "A-Team",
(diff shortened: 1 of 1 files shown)


barulicm:
Is this expected to fix the webkit load error? I'm still hitting that on my machine after running the deps install script and building the UI package from this branch.

chachmu:
try going into the `ateam_ui/src` folder and running `neu update --latest`. I thought upping the version number in the config file would handle that but I might need to tweak the setup script

barulicm (approved):

barulicm:
`--latest` was not recognized as an option, but just running `neu update` got it working.

chachmu:
That makes sense. I thought that the build script was already running `neu update` but it turns out it only does so if it doesn't detect an existing setup. I think its probably fine to leave it like this since we won't be updating neutralino versions very often.

PR #317, opened 2025-02-25
Fix kenobi world message data

### LINKED ISSUE #312: Kenobi World msg timestamp incorrect
I haven't had time to figure out why but there is definitely something wrong with the timestamp in the Kenobi world msg since the value always starts at 0 and increases very slowly, like 10 nanoseconds per second. Probably an issue in this line: https://github.com/SSL-A-Team/software/blob/ff5c5336290551dfa37de54c9e8ce65859f3407e/ateam_kenobi/src/types/message_conversions.cpp#L109-L112

chachmu:
Fixes several issues with the Kenobi Node world message:
* A double value of seconds was being passed into the nanoseconds field of the rclcpp:Time constructor
* The Ball and Robot toMsg conversion didn't populate the visible field
* Robots in the our and their robot msgs fields were only updated if they were "Available" which is probably not what we wanted for either option.

I did notice that Kenobi node currently seems to use a timestamp based off the time since starting your computer or something like that. Would it be better to swap to actual time since Unix epoch or something more absolute?

Closes #312

### DIFF
--- a/ateam_kenobi/src/kenobi_node.cpp
+++ b/ateam_kenobi/src/kenobi_node.cpp
@@ -259,6 +259,7 @@ private:
 
   void ball_state_callback(const ateam_msgs::msg::BallState::SharedPtr ball_state_msg)
   {
+    world_.ball.visible = ball_state_msg->visible;
     if (ball_state_msg->visible) {
       world_.ball.pos = ateam_geometry::Point(
         ball_state_msg->pose.position.x,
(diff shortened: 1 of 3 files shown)


barulicm (approved):

barulicm:
> I did notice that Kenobi node currently seems to use a timestamp based off the time since starting your computer or something like that. Would it be better to swap to actual time since Unix epoch or something more absolute?

Kenobi uses `std::chrono::steady_clock` for `current_time`. That clock source gives no real guarantees about when its start is, so "time since last reboot" is common.

There's maybe an interesting conversation to be had there about what clock we should use. `steady_clock` was chosen because the main use of that timestamp was to calculate time deltas between frames, and `steady_clock` ensures that will never deviate from actual elapsed time (ie no clock changes).

PR #323, opened 2025-03-10
Let goalie chip to clear the ball

barulicm:
This PR:
- Adds `KickType` to `PivotKick` and `LineKick` to allow specifying kick vs chip
- Adds check in `Goalie` skill's "clear ball" behavior to chip if the goalie robot has a chipper


ateam_kenobi/src/skills/goalie.cpp
chachmu, line 1: Long term it would probably be strategically valuable to allow the goalie to chip directly over opposing robots if it could get the ball into a more useful position but that will have to wait until we have a better understanding of our chippers dynamics

chachmu (approved):

PR #327, opened 2025-04-17
git ignore colcon dirs

CollinAvidano:


barulicm:
I'm inclined not to accept this change, but I'm open to discussion.
My main thought is that having those wrong folders under src/ can cause problems. (I don't have specifics on hand, but I remember this being an issue in the past.)
I think I'd rather have git try to yell at us / flag when we've accidentally built in the wrong directory rather than silently letting those bad folders exist.

barulicm:
Closing as stale

PR #350, opened 2025-07-04
Fixes usages of deprecated control_toolbox constructors.

barulicm:
The latest release of control_toolbox deprecated some constructors we were using for their objects to add their new `AntiWindupStrategy` system. This PR changes our code to use the new constructors.

### DIFF
--- a/ateam_kenobi/src/core/motion/motion_controller.cpp
+++ b/ateam_kenobi/src/core/motion/motion_controller.cpp
@@ -24,6 +24,7 @@
 #include <CGAL/squared_distance_2.h>
 #include <algorithm>
 #include <cmath>
+#include <limits>
 #include <vector>
 #include <ateam_msgs/msg/robot_motion_command.hpp>
 #include <ateam_msgs/msg/robot_state.hpp>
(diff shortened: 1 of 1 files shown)


ateam_kenobi/src/core/motion/motion_controller.cpp
```diff
@@ -195,10 +199,23 @@ ateam_msgs::msg::RobotMotionCommand MotionController::get_command(
 
 void MotionController::reset()
 {
-  // TODO(anon): handle pid gains better
-  this->x_controller.initialize(2.8, 0.0, 0.002, 0.3, -0.3, true);
-  this->y_controller.initialize(2.8, 0.0, 0.002, 0.15, -0.15, true);
-  this->t_controller.initialize(2.5, 0.0, 0.0, 0.5, -0.5, true);
+  const auto u_max = std::numeric_limits<double>::infinity();
```
chachmu, line 202: Would it be better to set these to like the max velocity or something just to make sure the controller doesn't go completely crazy?

chachmu (approved):

PR #354, opened 2025-07-07
motion controller overhaul with feedforward trapezoidal trajectories

chachmu:


ateam_geometry/src/intersection.cpp
```diff
@@ -74,6 +75,50 @@ std::optional<std::variant<Point, std::pair<Point, Point>>> intersection(
...
+  }
+
+  const auto intersection = maybe_intersection.value();
+
+  if (std::holds_alternative<Point>(intersection)) {
+    const auto intersection_point = std::get<Point>(intersection);
+
+    if (CGAL::do_intersect(segment, intersection_point)) {
+      return intersection_point;
+    }
```
barulicm, line 96: Does this approach handle the floating point mismatches well? It feels odd to use `do_intersect` in this branch and the distance check in the branch below.

barulicm (approved):

barulicm (approved):

PR #360, opened 2025-08-25
Comp 2025: radio packages

barulicm:
* Updates software_communication submodule
* Adds shut_down_robots and reboot_robots parameters to radio bridge
* Adds NaN protection to physical and simulation radio bridges
* Sets dribbler multiplier in control messages
* Improves log specificity in radio bridge


chachmu (approved):
I'm fine with things the way they are now but should the shutdown and reboot to it params actually be services instead?

barulicm:
> I'm fine with things the way they are now but should the shutdown and reboot robot params actually be services instead?

Agreed. #361

PR #370, opened 2025-09-01
Add motion benchmark package

barulicm:
This PR adds the current draft of our motion benchmark package, including the linear and angular velocity benchmarks.
More work to do in this area and to make running these benchmarks easier, but this version of the code was useful enough for us at comp that I think it's worth merging.


chachmu (approved):
I guess these could theoretically just subscribe to the specified robot number's topics since it is given that at startup but using the indexed subscriber code from Kenobi is probably easier

PR #379, opened 2026-03-04
New super vision filter

fourpenny:
A simpler vision filter that just uses basic Kalman filtering and that's about it. It keeps the same interface as our previous filter but has a slimmed down design.

You can test it:
`colcon test --packages-select new_vision_filter`

Or run the sim (or real) launch file:
`ros2 launch ateam_bringup bringup_simulation.launch.py`


new_super_vision_filter/package.xml
```diff
@@ -0,0 +1,24 @@
+<?xml version="1.0"?>
+<?xml-model href="http://download.ros.org/schema/package_format3.xsd" schematypens="http://www.w3.org/2001/XMLSchema"?>
+<package format="3">
+  <name>new_vision_filter</name>
```
barulicm: Package name should match the folder name

new_super_vision_filter/CMakeLists.txt
```diff
@@ -0,0 +1,80 @@
+cmake_minimum_required(VERSION 3.8)
+project(new_vision_filter)
+
+# if(CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID MATCHES "Clang")
+#   add_compile_options(-Wall -Wextra -Wpedantic)
+# endif()
```
barulicm: Consider turning these flags back on.

```suggestion
if(CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID MATCHES "Clang")
  add_compile_options(-Wall -Wextra -Wpedantic)
endif()
```

new_super_vision_filter/CMakeLists.txt
```diff
@@ -0,0 +1,80 @@
...
+if(BUILD_TESTING)
+  find_package(ament_cmake_lint_cmake REQUIRED)
+  find_package(ament_cmake_gtest REQUIRED)
+  # the following line skips the linter which checks for copyrights
+  # comment the line when a copyright and license is added to all source files
+  set(ament_cmake_copyright_FOUND TRUE)
+  # the following line skips cpplint (only works in a git repo)
+  # comment the line when this package is in a git repo and when
+  # a copyright and license is added to all source files
+  set(ament_cmake_cpplint_FOUND TRUE)
```
barulicm: Consider turning these linters back on.

```suggestion
```

new_super_vision_filter/src/measurements/ball_measurement.hpp
```diff
@@ -0,0 +1,44 @@
...
+
+class BallTrack {
+public:
+  BallTrack(const ssl_league_msgs::msg::VisionDetectionBall & ball_detection, int & camera_id)
+  : camera_id(camera_id)
+  {
+    PosMeasurement pos;
+    pos << ball_detection.pos.x,
+      ball_detection.pos.y;
+    timestamp = std::chrono::steady_clock::now();
```
barulicm, line 36: Should we think about using one of the timestamps from the vision detection frame message? Not sure if you've already considered this, and not sure if the time difference would actually matter in practice.

new_super_vision_filter/src/measurements/robot_measurement.hpp
```diff
@@ -0,0 +1,70 @@
...
+    ateam_common::TeamColor & team
+  )
+  : camera_id(camera_id), team(team)
+  {
+    PosMeasurement pos;
+    pos << bot_detection.pose.position.x,
+      bot_detection.pose.position.y;
+    angle << bot_detection.pose.orientation.w;
+    robot_id = bot_detection.robot_id;
+  }
```
barulicm, line 55: `timestamp` appears to not be initialized or set anywhere. Am I missing something?

new_super_vision_filter/src/measurements/ball_track.hpp
```diff
@@ -0,0 +1,44 @@
...
+// THE SOFTWARE.
+
+#ifndef BALL_TRACK_HPP_
+#define BALL_TRACK_HPP_
+
+#include "filter_types.hpp"
+
+#include <ssl_league_msgs/msg/vision_detection_ball.hpp>
+
+class BallTrack {
```
barulicm: Consider calling these types 'measurement' or 'detection'. At least in my mind, 'track' usually represents the internal filter state used to store detection history or correlate new detections to existing estimate instances.


```suggestion
class BallDetection {
```

new_super_vision_filter/src/filtered_ball.cpp
```diff
@@ -0,0 +1,73 @@
...
+
+void FilteredBall::update(const BallTrack & track)
+{
+    // Make sure this detection isn't crazy off from our previous ones
+    // (unless our filter is still new/only has a few measurements)
+  if (age < oldEnough) {
+    ++age;
+  }
+  if (health < maxHealth) {
+    health += 2;
```
barulicm, line 41: Does `health` just track the staleness of the ball detections? Seems like it might be useful to include some measure of filter confidence in that metric as well.

new_super_vision_filter/src/vision_filter_node.cpp
```diff
@@ -0,0 +1,264 @@
...
+
+                // Process all the new updates from the tracks
+      for (const auto & bot_track : blue_tracks) {
+        ateam_common::TeamColor team_color = ateam_common::TeamColor::Blue;
+        auto it = std::find_if(
+                        blue_robots.begin(),
+                        blue_robots.end(),
+          [bot_track](const FilteredRobot & bot){return bot_track.getId() == bot.getId();}
+        );
+        if (it != blue_robots.end()) {
```
barulicm: Is this condition inverted? It looks like you're trying to create `FilteredRobot` instances on demand as new IDs pop up. `iterator != end` means that an existing element was found.

This should also segfault as written when it takes the `else` branch as `it` does not point to a valid object.

barulicm (commented):
Some interesting stuff in here. Excited to see a fresh start to this.
We should add more tests to fully exercise as much of the code and math as possible.
We may want to have a deeper conversation about architecture and object management in this at some point. There might be some simplifications we can make, especially on the robot tracking side where we know exactly how many robots we could be tracking.

new_super_vision_filter/package.xml
fourpenny: Do you have a preference for which we should pick? I'll change them all to match.

new_super_vision_filter/src/measurements/ball_measurement.hpp
fourpenny: Hm, I think it depends on whether we think degradation in quality/lag might be more likely from our vision filter internally or from SSL vision. Do you think it would be worth checking both? I think that switching to the SSL time stamps or adding them as an additional quality check would be straightforward.

new_super_vision_filter/src/measurements/robot_measurement.hpp
fourpenny: This should be initialized now in the constructor for the RobotMeasurement!

new_super_vision_filter/src/filtered_ball.cpp
fourpenny: For now it just checks the staleness. We could make this number a combination of multiple checks, or another idea I have is to rename this something related to its function ("staleness", maybe) and then add other explicit defined checks related to the filter quality. My thought is that's a good option, since we can continue adding more as we go for now, until we have a better idea of what the best quality checks are.

new_super_vision_filter/src/vision_filter_node.cpp
fourpenny: Whoops... this is fixed. I'm impressed (I think?) it never segfaulted in testing...

new_super_vision_filter/src/measurements/ball_track.hpp
fourpenny: Changed both this and the RobotTrack class to use the suffix "Measurement" instead.

new_super_vision_filter/package.xml
fourpenny: I made it "new_super_vision_filter" for now, since that's more fun, but if the length is a problem I can change it to "new_vision_filter".

fourpenny:
> Some interesting stuff in here. Excited to see a fresh start to this. We should add more tests to fully exercise as much of the code and math as possible. We may want to have a deeper conversation about architecture and object management in this at some point. There might be some simplifications we can make, especially on the robot tracking side where we know exactly how many robots we could be tracking.

I think at least as an initial test before the PTO this weekend, I'll plan to see what generally happens to the filter state in a few conditions we're likely to see:
- If we have multiple balls and/or a very noisy ball
- If we have robots dropping in and out of visibility
- If we have two of the same robot on the field (like if we practice on a field with another team with Yellow Robot 1)... I think this is more of a stretch goal/we can rush to address it if needed.

new_super_vision_filter/src/measurements/ball_measurement.hpp
barulicm: This might just be something we take a deeper look at later with more real world testing. One complicating factor is that we are not actually time-syncing with the vision servers, so their timestamps might not be valid on our machine.

PR #384, opened 2026-04-01
GC Launch Revamp

barulicm:
* Adds support for native GC install
* Removes dependency on net-tools by using the `-address` flag
* Reorganizes README a bit for clarity

To support the new behaviors, I had to switch the GC launch file to Python.

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,7 +42,7 @@
    # In the ateam_ws directory
    ./src/software/ateam_ui/install_deps.sh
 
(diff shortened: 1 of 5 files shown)


chachmu (approved):

barulicm:
We noticed tonight that bringup_simulation restricts the GC bridge nodes to use the docker network interface. This means running this launch file with the native GC will not work.
Should give this some thought before merging.

PR #389, opened 2026-04-28
Radio bridge connection states

barulicm:
The motivating change here was adding the initial connection timeout feature that @guyfleeman had prototyped. Rather than continue to accrue boolean arrays that track the connection lifecycle going forward, I've added the `ConnectionState` enum where we can add as many states as we want and keep it in one array.

While I was poking around, I decided to update the launch tests in this package so they would catch bugs like the goodbye message bug fixed in #385 . I also turned on warnings as errors for this package as a prep step for the more substantial changes coming to support the new control scheme. Oh, and I made the plural-ish-ness of some variable names more consistent.


guyfleeman:
Seems good to me. Should lay the ground work for time sync as well if we need that at some point.

guyfleeman:
I'm okay with the code. Do you wanna test on main first or nah?

barulicm:
> I'm okay with the code. Do you wanna test on main first or nah?

I'm confident enough especially with the improved launch tests passing.

chachmu (approved):

PR #395, opened 2026-05-17
Resets onboard controller during side changes

barulicm:
This PR updates the communications repo and adds a logic to hold `reset_controller` high during field side changes.


radio/ateam_radio_bridge/src/radio_bridge_node.cpp
```diff
@@ -260,11 +262,12 @@ class RadioBridgeNode : public rclcpp::Node
   void SendCommandsCallback()
   {
     const std::lock_guard lock(mutex_);
+    const auto now = std::chrono::steady_clock::now();
```
chachmu, line 265: Did we ever standardize on using chrono or ros clock for our timestamps everywhere?

chachmu (approved):

radio/ateam_radio_bridge/src/radio_bridge_node.cpp
barulicm: Not sure if we ever formally decided, but right now we only use chrono everywhere.

PR #397, opened 2026-05-22
zero init radio bridge packets

nickwitten:
Zero initialize radio packets so that garbage isn't sent in fields that aren't explicitly set.


chachmu (approved):

chachmu:
My understanding is that ROS Jazzy already initializes the data in the message type even without using braces, but it doesn't hurt to be explicit about it. Are you actually seeing issues caused by this?

barulicm:
> My understanding is that ROS Jazzy already initializes the data in the message type even without using braces, but it doesn't hurt to be explicit about it. Are you actually seeing issues caused by this?

This PR is initializing the radio packet struct, which is not a ROS message type.

barulicm (approved):

chachmu:
🤦 I keep forgetting the radio bridge internal types aren't ROS messages

PR #433, opened 2026-07-09
Comp merge: ateam_kenobi STP

barulicm:
Imports changes to our skills, tactics, and plays.

### DIFF
--- a/ateam_kenobi/src/skills/goalie.cpp
+++ b/ateam_kenobi/src/skills/goalie.cpp
@@ -86,16 +86,20 @@ void Goalie::runFrame(
   const auto time_in_def_area = ateam_common::TimeDiffSeconds(world.current_time,
       ball_entered_def_area_time_);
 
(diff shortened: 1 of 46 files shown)


ateam_kenobi/src/skills/goalie.cpp
chachmu, line 1: Goalie might need a state for starting a penalty where it stays far enough back to be legal

ateam_kenobi/src/skills/pass_receiver.cpp
chachmu, line 1: I think during the last match we saw that having the backup time set to be 1 second was way too long

ateam_kenobi/src/tactics/blockers.cpp
chachmu, line 1: We might want to do something smarter if our blockers are being sent to the other half of the field

chachmu (approved):

